PR #9, opened 2020-06-08
PayPal method is always shown as first.

SirDomin:


maximehuran:
Why do we have Paypal in first ?

PR #129, opened 2020-10-02
README improvements

Zales0123:
Without this change, the plugin is installing by default in `0.3.1` version, which is quite old 💃


armellarcier:
@Zales0123 why did you declare a conflict with "symplify/package-builder": "^8.3.24" ?

I'm having issues installing the plugin, see #173

Issue #115, opened 2020-09-23
Allow user to onboard only one seller


SirDomin:
Enable PayPal payment method should be hidden when we got already onboarded seller, having 2 sellers may cause problems.
![image](https://user-images.githubusercontent.com/22825722/94021096-f1726200-fdb3-11ea-8a58-59fb4141a345.png)

PR #136, opened 2020-10-13
Allow only one seller to be onboarded

### LINKED ISSUE #126: Problem with webhooks when 2 paypal payment methods are in the system
Api requests are authorized with first paypal payment method as we dont get any id's from first webhook. So when we got 2 paypal payment methods we cant authorize refunds from second payment.

Zales0123:
Fixes #126 and #115

Even though ultimately we would like to allow one PayPal payment method for each channel, right now multiple PayPal PMs can lead to some unexpected behaviour (as in linked issues). I propose to allow only one seller until the stable release and fix the problems in the later plugin versions 💃


SirDomin:
Thanks, Mateusz! :1st_place_medal:

PR #148, opened 2020-10-19
Do not assign PayPal customer to order if the user is logged in

### LINKED ISSUE #137: The order isn't associated to the connected user
When I perform a payment, the order is associated to the PayPal account's email, not to my account (I was connected).

The order should be associated to the connected user.

Zales0123:
Fixes #137 

Logged in user is automatically set as an order's customer 💃


SirDomin:
Thanks, Mateusz! :tada:

PR #150, opened 2020-10-26
Single page payment

### LINKED ISSUE #116: Single page payment page
To improve user experience payment page should be done without redirecting.
image

SirDomin:
image
instead of default sylius button we get paypal form to finish our payment
fixes https://github.com/Sylius/PayPalPlugin/issues/116


src/Controller/PayPalButtonsController.php
```diff
@@ -129,4 +144,39 @@ public function renderPaymentPageButtonsAction(Request $request): Response
...
+    {
+        $orderId = $request->attributes->getInt('orderId');
+        /** @var OrderInterface $order */
+        $order = $this->orderRepository->find($orderId);
+        /** @var PaymentInterface $payment */
+        $payment = $order->getLastPayment();
+        /** @var PaymentMethodInterface $paymentMethod */
+        $paymentMethod = $payment->getMethod();
+        /** @var GatewayConfigInterface $gatewayConfig */
+        $gatewayConfig = $paymentMethod->getGatewayConfig();
```
Zales0123, line 158: There are so many places where we get data from the gateway config :/ We should definitely extract some service for that (but, of course, not in this PR 🖖)

Zales0123 (changes requested):
We need some more tests on that :/ 

- after successful payment, there is another, _cart_ payment in the fulfilled order

- when you try to pay by PayPal or hosted payment method (like Przelewy24) and cancel it, it returns to complete page with processing payment - and after clicking "Place order" there is a `"Unable to generate a URL for the named route "sylius_paypal_plugin_pay_with_paypal_form" as such route does not exist.` exception.

src/Resources/views/bundles/SyliusShopBundle/Checkout/complete.html.twig
```diff
@@ -0,0 +1,35 @@
...
+
+        <div class="ui hidden divider"></div>
+        {% include '@SyliusShop/Checkout/Complete/_form.html.twig' %}
+
+        {{ sylius_template_event('sylius.shop.checkout.complete.before_navigation', {'order': order}) }}
+
+        {% if order.lastPayment.method.gatewayConfig.factoryName == 'sylius.pay_pal' and order.lastPayment.state == 'cart' %}
+            {{ render(controller('Sylius\\PayPalPlugin\\Controller\\PayPalButtonsController:renderPayPalPaymentAction', {'orderId': order.id})) }}
+        {% else %}
+            {% include '@SyliusShop/Checkout/Complete/_navigation.html.twig' %}
```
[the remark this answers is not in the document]
SirDomin: right, one template is better :D

Zales0123:
I've pushed some visual changes, which I think would be more consistent with the default "Place order" UI:

Zrzut ekranu 2020-10-29 o 09 24 19
Zrzut ekranu 2020-10-29 o 09 24 29

Zales0123:
There is still a problem with duplicated payments after the successful PayPal payment :/ 

Zrzut ekranu 2020-10-29 o 09 16 12

It would be nice to fix it. I would also like to merge it after the 1.0.0 release, as it's the improvement/change of the currently working feature, that could quite possibly change a lot :)

src/Controller/CompletePayPalOrderAction.php
```diff
@@ -23,25 +26,42 @@ final class CompletePayPalOrderAction
...
         $this->orderProvider = $orderProvider;
+        $this->stateMachineFactory = $stateMachineFactory;
+        $this->orderManager = $orderManager;
     }
 
     public function __invoke(Request $request): Response
     {
-        $token = (string) $request->attributes->get('token');
-        $order = $this->orderProvider->provideOrderByToken($token);
+        $id = (int) $request->attributes->get('id');
```
jacquesbh, line 51: I think using the `id` could lead to a flow issue here. If I see that in my network, I will try to do the same with `+1` and `+2` and so on.

So, maybe we could stick to the token?

Actually a much better way is to have `token` and `id`, so it's almost impossible to guess.

src/Controller/PayPalButtonsController.php
```diff
@@ -135,4 +150,39 @@ public function renderPaymentPageButtonsAction(Request $request): Response
             return new Response('');
         }
     }
+
+    public function renderPayPalPaymentAction(Request $request): Response
```
jacquesbh, line 154: I'm not a huge fan of the "no verification at all" to who is the owner of the order. Changing the `orderId` in the request could lead to some issues here. Is it right?

jacquesbh (changes requested):

src/Controller/CompletePayPalOrderAction.php
Zales0123: Ok, we cannot use the token, because it's not generated yet. *But* we can use a PayPal order ID which is also quite unique, so would be much safer than the autogenerated ID :)

NoResponseMate:
Hi @SirDomin 👋 

Although the topic at hand is very much a 👍, I'm closing this as it's just too heavily outdated (contains BC breaks, a few other issues 💃).

We'll make a second attempt at this functionality, based on the currently supported versions.

Cheers 🍻

Issue #133, opened 2020-10-12
Log error when an error occured while on PayPal


LucaGallinari:
In the PayPalPaymentOnErrorAction class there is a todo regarding this:
https://github.com/Sylius/PayPalPlugin/blob/v1.0.0-RC.1/src/Controller/PayPalPaymentOnErrorAction.php#L23

This is really important cause is not trivial to debug without some logging.

Zales0123:
Agreed 👍 The good thing is we already have the monolog with the PayPal channel configured, so it should be doable quite easily :)

PR #151, opened 2020-10-26
log payments error when onError triggers

SirDomin:
image
example error
fixes https://github.com/Sylius/PayPalPlugin/issues/133


Zales0123 (approved):
We probably should check it twice to be 100% sure that each error would be handled properly - but it's still better than writing no log at all 🖖

Zales0123:
Thanks, @SirDomin! :tada:

Issue #131, opened 2020-10-05
Some locales are not supported on PayPal


LucaGallinari:
I configured the payment method with a sandbox account. After clicking on the place order button in the checkout, i land on the payment page but the paypal form/buttons are missing (instead the credit card tab is filled): 

<img width="500" alt="payment" src="https://user-images.githubusercontent.com/5822554/95053690-0ffc2580-06f1-11eb-91d9-b0b50580778d.png">

And in the console i found this:

> GET https://www.paypal.com/sdk/js?components=hosted-fields,buttons&locale=it&client-id=****&merchant-id=****&intent=capture net::ERR_ABORTED 400

If I call the URL i get this error:

> "Invalid query value for locale: it"

If i call it with it_IT locale it works.
As i can see from the supported locales in the Paypal docs (https://developer.paypal.com/docs/api/reference/locale-codes/#supported-locale-codes) they do not support 2-digit locales. 

So probably a mapping between Sylius locales and Paypal locales is needed?

Zales0123:
Hi @LucaGallinari! What prevents you from setting the locale to `it_IT`? It's still the recommended way to do it in Sylius. I understand, that having a regular `it` selectable in the locales list can be confusing, but it's provided by the PHP intl component so we have no control over it.

Still, some mechanism that would map 2-digit locales to their extended, default versions (`it` -> `it_IT`, `pl` -> `pl_PL` etc.) would be definitely an improvement 👌

LucaGallinari:
Hi @Zales0123 sorry for the late response. We used "it" cause this locale is bound to the "it" channel in which you can ship to Italy (it_IT), San Marino (it_SM) and Città del Vaticano (it_VA). So the resulting URL will be mywebsite.com/it and not mywebsite.com/it_IT cause this one would be misleading, we don't want to depend on the region. Moreover, from a SEO perspective is better, cause the language is a generic "italian" not bound to the region.

PR #152, opened 2020-10-26
Process locales to match paypal requirements

SirDomin:
fixes https://github.com/Sylius/PayPalPlugin/issues/131

### DIFF
--- a/spec/Processor/LocaleProcessorSpec.php
+++ b/spec/Processor/LocaleProcessorSpec.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace spec\Sylius\PayPalPlugin\Processor;
+
+use PhpSpec\ObjectBehavior;
+
+final class LocaleProcessorSpec extends ObjectBehavior
+{
+    public function it_processes_locale_to_match_locale_regex(): void
+    {
+        $this->process('uk')->shouldReturn('uk_UK');
+        $this->process('pl')->shouldReturn('pl_PL');
+    }
+
+    public function it_returns_same_locale_if_already_matches_regex(): void
+    {
+        $this->process('it_IT')->shouldReturn('it_IT');
+    }
+}
--- a/spec/Processor/LocaleProcessorSpec.php
+++ b/spec/Processor/LocaleProcessorSpec.php
@@ -10,8 +10,9 @@ final class LocaleProcessorSpec extends ObjectBehavior
 {
     public function it_processes_locale_to_match_locale_regex(): void
     {
-        $this->process('uk')->shouldReturn('uk_UK');
+        $this->process('et')->shouldReturn('et_EE');
         $this->process('pl')->shouldReturn('pl_PL');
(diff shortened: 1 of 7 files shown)


spec/Processor/LocaleProcessorSpec.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace spec\Sylius\PayPalPlugin\Processor;
+
+use PhpSpec\ObjectBehavior;
+
+final class LocaleProcessorSpec extends ObjectBehavior
+{
+    public function it_processes_locale_to_match_locale_regex(): void
+    {
+        $this->process('uk')->shouldReturn('uk_UK');
```
Zales0123, line 13: There is no such locale as `uk` 😄 there is `en_GB` 💃

Zales0123 (changes requested):
It will not work properly :/ For less common languages (like Polish) the country code will be indeed the uppercase of locale code (`pl` -> `pl_PL`). However, for the most common one (English) it will result in `en` -> `en_EN` (should be `en_US` or `en_GB`), not even mentioning cases like `ja` -> `ja_JA` (should be `ja_JP`) or `et` -> `et_ET` (should be `et_EE`).

I think that the most straightforward (even if not elegant) way, would be to create a hardcoded array of default regions for languages. As far as I see, Symfony [Intl component](https://symfony.com/doc/current/components/intl.html#locales) provides 136 locales without region specified (we can use `Locales::getLocales()` for that). It would be... coarse but should work in all cases 🖖 🚀

src/Processor/LocaleProcessor.php
```diff
@@ -0,0 +1,30 @@
...
+            return $locale;
+        }
+
+        $locales = Locales::getLocales();
+        $localeFound = (int) array_search($locale, $locales);
+        while (!$this->isValidLocale($locales[$localeFound]) && $localeFound < count($locales)) {
+            ++$localeFound;
+        }
+
+        return $locales[$localeFound];
```
Zales0123, line 23: I'm still not 100% sure about this implementation :D indeed, it will not result in invalid locales (which is definitely an improvement), but it can result in very exotic default locales, like `fr_BE` for `fr`, `en_AE` for `en` and `de_AT` for `de` :/ 

I would still go with a hardcoded array of default locales. The alternative way would be an algorithm like:

1. Get the default locale from the array of the most common locales (like `en_US`, `fr_FR`, `de_DE` etc.)
2. If not, try to get the locale with the region being an uppercase version of locale (like it was before: `pl` -> `pl_PL`)
3. If not, get the locale with **this** implementation 

But IMO it's a little bit overcomplication for such a simple case :) 💃

Zales0123: OK, I answer myself - still, the recommended configuration is to use the locale *with* region, so even if the service provides some exotic locale... it's only the default value :) I believe we will be able to fix it if it results in some problems 🖖

Zales0123 (approved):

Zales0123:
Thanks, @SirDomin! :1st_place_medal:

PR #153, opened 2020-10-29
Inform about empty billing address after PayPal checkout

### LINKED ISSUE #139: [Feedback] Empty addresses after payment?
After creating an account from a guest order (using the PayPal button on the product page directly), it seems that I have two addresses in my account, but they are empty:

image

Zales0123:
It maybe not fixes #139, but somehow handles the problem. We indeed have no place to get the billing address so it should be passed manually in the addressing step (or should be taken from the default customer address).

Zrzut ekranu 2020-10-29 o 23 44 22

### DIFF
--- a/src/Resources/views/bundles/SyliusShopBundle/Common/Order/_addresses.html.twig
+++ b/src/Resources/views/bundles/SyliusShopBundle/Common/Order/_addresses.html.twig
@@ -1,0 +1,23 @@
+<div class="ui segment">
+    <div class="ui {% if order.isShippingRequired() %}two{% else %}one{% endif %} column divided stackable grid">
+        <div class="column" id="sylius-billing-address" {{ sylius_test_html_attribute('billing-address') }}>
+            <div class="ui small dividing header">{{ 'sylius.ui.billing_address'|trans }}</div>
+            {% include '@SyliusShop/Common/_address.html.twig' with {'address': order.billingAddress} %}
+        </div>
+        {% if order.isShippingRequired() %}
+        <div class="column" id="sylius-shipping-address" {{ sylius_test_html_attribute('shipping-address') }}>
+            <div class="ui small dividing header">{{ 'sylius.ui.shipping_address'|trans }}</div>
+            {% include '@SyliusShop/Common/_address.html.twig' with {'address': order.shippingAddress} %}
+        </div>
+        {% endif %}
+    </div>
+</div>
+{% if not order.isShippingRequired() and order.billingAddress.street == '' %}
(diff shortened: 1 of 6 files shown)


src/Resources/views/bundles/SyliusShopBundle/Common/Order/_addresses.html.twig
```diff
@@ -0,0 +1,23 @@
...
+        </div>
+        {% if order.isShippingRequired() %}
+        <div class="column" id="sylius-shipping-address" {{ sylius_test_html_attribute('shipping-address') }}>
+            <div class="ui small dividing header">{{ 'sylius.ui.shipping_address'|trans }}</div>
+            {% include '@SyliusShop/Common/_address.html.twig' with {'address': order.shippingAddress} %}
+        </div>
+        {% endif %}
+    </div>
+</div>
+{% if not order.isShippingRequired() and order.billingAddress.street == '' %}
```
SirDomin, line 15: why we check only street here? maybe twig boolean function would be better, also implement there to be sure everything is empty?

src/Resources/views/bundles/SyliusShopBundle/Common/Order/_addresses.html.twig
```diff
@@ -0,0 +1,23 @@
+<div class="ui segment">
+    <div class="ui {% if order.isShippingRequired() %}two{% else %}one{% endif %} column divided stackable grid">
```
SirDomin, line 2: maybe if we added block from line https://github.com/Sylius/PayPalPlugin/pull/153/files#diff-1d7e4473bdf0d61d613efa015fe5c9998422a98086d812a6a7868cff88a2e6d0R15 it would look better to display info about missing address instead of comma.
image

What do you think?

Zales0123: Agreed 👍 However, it would require from us extending another template: `@SyliusShop/Common/_address.html.twig`, which will be also another step of the plugin's installation. I would vote to fix it in Sylius itself (not displaying the comma if there is no street in the address) 🖖

src/Resources/views/bundles/SyliusShopBundle/Common/Order/_addresses.html.twig
Zales0123: I will create a twig function, checking city and postcode as well 👌

SirDomin:
Thanks, Mateusz! :1st_place_medal:

PR #158, opened 2020-11-11
Check order has a shipping address in the PayPalAddressProcessor

### LINKED ISSUE #157: Take in account orders without shipping address
For example for products that are digital subscriptions that don't need to be shipped

Actually [PayPalAddressProcessor](https://github.com/Sylius/PayPalPlugin/blob/master/src/Processor/PayPalAddressProcessor.php) fails if order hasn't a shipping address.

jdeveloper:
Fixes #157

### DIFF
--- a/src/Processor/PayPalAddressProcessor.php
+++ b/src/Processor/PayPalAddressProcessor.php
@@ -27,7 +27,7 @@ final class PayPalAddressProcessor implements PayPalAddressProcessorInterface
         /** @var ?AddressInterface $orderAddress */
         $orderAddress = $order->getShippingAddress();
 
-        if (null == $orderAddress) {
+        if (null === $orderAddress) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src/Processor/PayPalAddressProcessor.php
```diff
@@ -24,9 +24,13 @@ public function __construct(ObjectManager $objectManager)
      */
     public function process(array $address, OrderInterface $order): void
     {
-        /** @var AddressInterface $orderAddress */
+        /** @var ?AddressInterface $orderAddress */
         $orderAddress = $order->getShippingAddress();
 
+        if (null == $orderAddress) {
```
Zales0123, line 30: We should always use a strict comparison `===`

Zales0123 (commented):
👍

Zales0123 (approved):

Zales0123:
Thanks, @jdeveloper! :1st_place_medal:

PR #159, opened 2020-11-12
Require Sylius ^1.8

Zales0123:
As Sylius 1.7 is coming to the end (as explained in [The Release Cycle](https://docs.sylius.com/en/latest/book/organization/release-cycle.html) documentation section), I would like to require Sylius 1.8 in this plugin. It also simplifies the test pipelines a lot.

I think that we can still introduce some critical bug fixes (if any appears) to 1.0.* version for some time (as I can imagine not everyone can move to the latest version of Sylius - even though I strongly encourage to do that 😄).


lchrusciel (approved):

GSadee (approved):

Tomanhez (approved):

.github/workflows/build.yml
```diff
@@ -51,22 +51,6 @@ jobs:
                 with:
                     node-version: "${{ matrix.node }}"
 
-            -
-                name: Require Sylius
-                run: composer require sylius/sylius:${{ matrix.sylius }} --no-interaction --no-update
```
pamil, line 56: If this step gets deleted, then setting Sylius version in the matrix makes no sense - let's either keep it or remove Sylius version from the matrix.

SirDomin:
Thank you, Mateusz! :1st_place_medal:

PR #161, opened 2020-11-13
[Migration] Adjust namespace to newest Sylius Standard

lchrusciel:
Despite the note: `Namespace is arbitrary but should be different from App\Migrations as migrations classes should NOT be autoloaded`, the current Sylius Standard configuration suggests `App\Migration` as a namespace(ref. https://github.com/Sylius/Sylius-Standard/blob/master/config/packages/doctrine_migrations.yaml#L6). What is more, this namespace is not an issue anymore: https://github.com/doctrine/DoctrineMigrationsBundle/issues/242.

Without this change, current migration will not work automatically during plugin installation.

### DIFF
--- a/etc/travis/sylius18/doctrine_migrations.yaml
+++ b/etc/travis/sylius18/doctrine_migrations.yaml
@@ -3,4 +3,4 @@ doctrine_migrations:
         table_storage:
             table_name: sylius_migrations
     migrations_paths:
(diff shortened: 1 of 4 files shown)


Zales0123 (approved):

Zales0123:
Thank you, Łukasz! :tada:

PR #166, opened 2020-11-16
[Payum] Use custom resolve next route action only for its own gateway

lchrusciel:

PR #167, opened 2020-11-16
[Payum] Define resolve next route action only for paypal gateway

lchrusciel:
Replacement of #166

### DIFF
--- a/src/Resources/config/services/payum.xml
+++ b/src/Resources/config/services/payum.xml
@@ -34,7 +34,7 @@
         </service>
 
         <service id="Sylius\PayPalPlugin\Payum\Action\ResolveNextRouteAction" public="true">
-            <tag name="payum.action" all="true" alias="sylius.resolve_next_route" />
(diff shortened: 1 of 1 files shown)


src/Payum/Action/ResolveNextRouteAction.php
```diff
@@ -41,20 +39,6 @@ public function execute($request): void
...
-        }
-
-        /** @var PaymentInterface $model */
-        $model = $request->getFirstModel();
-        /** @var PaymentMethodInterface $paymentMethod */
-        $paymentMethod = $model->getMethod();
-        /** @var GatewayConfigInterface $gatewayConfig */
-        $gatewayConfig = $paymentMethod->getGatewayConfig();
-
-        return $gatewayConfig->getFactoryName() === 'sylius.pay_pal';
```
Zales0123, line 58: I know it's handled in the configuration, but maybe we should leave this check as it is? We cannot be *too much* secured and still somebody can override the service definition and change the tag 💃

Zales0123 (approved):

Zales0123:
Thanks, Łukasz! :tada:

PR #170, opened 2020-11-28
 API Platform support

SirDomin:
By using plugin class we can access its data by request from ApiPlatform within Sylius api platform payment integration.
Some refactor is needed inside PayPalPlugin.

### DIFF
--- a/src/ApiPlatform/PayPalPayment.php
+++ b/src/ApiPlatform/PayPalPayment.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\PayPalPlugin\ApiPlatform;
+
+use Sylius\Bundle\PayumBundle\Model\GatewayConfigInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\PaymentInterface;
+use Sylius\Component\Core\Model\PaymentMethodInterface;
+use Sylius\PayPalPlugin\Provider\AvailableCountriesProviderInterface;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+//TODO: once we implement ApiPlatformPayment interface in core, implement it here
+final class PayPalPayment
+{
+    /** @var RouterInterface */
+    private $router;
+
+    /** @var AvailableCountriesProviderInterface */
+    private $availableCountriesProvider;
+
+    public function __construct(RouterInterface $router, AvailableCountriesProviderInterface $availableCountriesProvider)
+    {
+        $this->router = $router;
+        $this->availableCountriesProvider = $availableCountriesProvider;
+    }
+
+    public function supports(PaymentMethodInterface $paymentMethod): bool
+    {
+        /** @var GatewayConfigInterface $gatewayConfig */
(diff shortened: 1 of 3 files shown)


src/ApiPlatform/PayPalPayment.php
```diff
@@ -0,0 +1,81 @@
...
+            'completePayPalOrderFromPaymentPageUrl' => $this->router->generate(
+                    'sylius_paypal_plugin_complete_paypal_order',
+                    ['token' => $order->getTokenValue()],
+                    UrlGeneratorInterface::ABSOLUTE_URL
+                ),
+            'createPayPalOrderFromPaymentPageUrl' => $this->router->generate(
+                    'sylius_paypal_plugin_create_paypal_order',
+                    ['token' => $order->getTokenValue()],
+                    UrlGeneratorInterface::ABSOLUTE_URL
+                ),
```
Zales0123, line 70: Here as well

src/ApiPlatform/PayPalPayment.php
```diff
@@ -0,0 +1,81 @@
...
+
+use Sylius\Bundle\PayumBundle\Model\GatewayConfigInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\PaymentInterface;
+use Sylius\Component\Core\Model\PaymentMethodInterface;
+use Sylius\PayPalPlugin\Provider\AvailableCountriesProviderInterface;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+//TODO: once we implement ApiPlatformPayment interface in core, implement it here
```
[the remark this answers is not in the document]
Zales0123: And I would either wait for it or (better) implement a custom interface here and then extend the mentioned one in it

src/ApiPlatform/PayPalPayment.php
```diff
@@ -0,0 +1,81 @@
...
+final class PayPalPayment
```
Zales0123, line 25: We're also missing some spec or test for this class

src/Resources/config/services.xml
```diff
@@ -26,6 +26,12 @@
             class="Sylius\PayPalPlugin\Generator\PayPalAuthAssertionGenerator"
         />
 
+        <service id="Sylius\PayPalPlugin\ApiPlatform\PayPalPayment">
+            <argument type="service" id="router" />
+            <argument type="service" id="Sylius\PayPalPlugin\Provider\AvailableCountriesProviderInterface" />
+            <tag name="api.payment" />
```
Zales0123, line 32: Can you explain what does this tag do?

Zales0123 (commented):
I feel a little bit worried that we're introducing something that has no use in the plugin itself :( Can we at least test it with some integration/functional test?

src/Resources/config/services.xml
SirDomin: this tag is used to collect all data providers from payment plugins, in Core we got PaymentConfiguration provider which allows us to send data to frontend based on picked payment method, f.eg if we pick paypal payment method we search throughout all payment methods with api.payment tag and see which one supports it, then we send payment data configuration to pay using specific payment method.

src/ApiPlatform/PayPalPayment.php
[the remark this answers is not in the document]
SirDomin: there will be implemented custom interface which will allow us to remove api.payment tag from services in PayPalPlugin, it all will be loaded automatically once we implement this interface in Core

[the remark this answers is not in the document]
SirDomin: https://github.com/Sylius/Sylius/pull/12107

Zales0123 (approved):

Zales0123:
Thanks, @SirDomin! :1st_place_medal:

PR #172, opened 2020-12-02
Adding new payment method fix

tifed:


Zales0123:
Thank you for the fix 🖖

It would also nice to express it in [this scenario](https://github.com/Sylius/PayPalPlugin/blob/master/features/trying_to_onboard_more_than_one_pay_pal_seller.feature), as it only represents the situation with PayPal already onboarded. We would need one scenario checking do get the error with PayPal seller already onboarded (the one we have) and the other one, checking can we create an offline PM. Do you think you would be able to give it a try?

Also, it would be great if you could squash your commits, as 4 of them for such a small change is definitely too much 🖖 :) 

Thank you one more time!

Zales0123:
The base of this pull-request was changed, you need fetch and reset your local branch 
if you want to add new commits to this pull request. **Reset before you pull, else commits
may become messed-up.**

Unless you added new commits (to this branch) locally that you did not push yet, 
execute `git fetch origin && git reset "master"` to update your local branch.
 
Feel free to ask for assistance when you get stuck :+1:

Zales0123:
Thanks, @tifed! :tada:

PR #179, opened 2020-12-22
Use sylius twig filter to get the proper province name

LucaGallinari:
This way the field will always be populated with the province name, even if the address contains only the province code.


Zales0123 (approved):
Hello @LucaGallinari, it's a long time 😄 The problem I have with this change is that it would result in a database call in some situations... On the other hand, there usually will be a province code or name, so it's just protection...

Nevermind, it's good to go 🚀

Zales0123:
Thank you, Luca! :tada:

PR #190, opened 2021-04-01
Bug #183 - Fix totals decimal precision in PayPalItemDataProvider

### LINKED ISSUE #183: Paypal Order creation fails due to wrong decimal precision
Sometimes calls to https://api.paypal.com/v2/checkout/orders return a 422 status code response with the following content:
```
Array(
    [name] => UNPROCESSABLE_ENTITY
    [details] => Array(
            [0] => Array(
                    [field] => /purchase_units/@reference_id=='XXX'/amount/value
                    [value] => 242.79999999999998
                    [issue] => DECIMAL_PRECISION
                    [description] => If the currency supports decimals, only two decimal place precision is supported.
                )
        )
    [message] => The requested action could not be performed, semantically incorrect, or failed business validation.
    [debug_id] => XXX
    [links] => Array(
            [0] => Array(
                    [href] => https://developer.paypal.com/docs/api/orders/v2/#error-DECIMAL_PRECISION
                    [rel] => information_link
                    [method] => GET
                )
        )
)
```

This is due to the fact that decimal precision is not properly enforced before calling the Paypal Create Order API: it should have exactly 2 decimals. Prices/amounts stored in the DB could be divided by 100 and you should always obtain a number with 2 decimals but the PayPalItemDataProvider does some sums and multiplications on the amounts and it lead to a floating point error.

This could be solved in CreateOrderApi class by enforcing decimal for all amount while setting up data, with something like this:
`number_format($amount, 2, '.', '')`

---------------

Moreover, but maybe it could be discussed in a separate issue, Paypal seems to require a different decimal precision based on the currency: https://developer.paypal.com/docs/api/reference/currency-codes/

jeandanyel:
Fix #183

### DIFF
--- a/src/Provider/PayPalItemDataProvider.php
+++ b/src/Provider/PayPalItemDataProvider.php
@@ -36,8 +36,8 @@ final class PayPalItemDataProvider implements PayPalItemDataProviderInterface
             foreach ($nonNeutralTaxes as $nonNeutralTax) {
                 $displayQuantity = $nonNeutralTaxes === [0] ? $orderItem->getQuantity() : 1;
                 $itemValue = $orderItem->getUnitPrice();
-                $itemData['total_item_value'] += ($itemValue * $displayQuantity) / 100;
-                $itemData['total_tax'] += ($nonNeutralTax * $displayQuantity) / 100;
+                $itemData['total_item_value'] += $itemValue * $displayQuantity;
+                $itemData['total_tax'] += $nonNeutralTax * $displayQuantity;
 
                 $itemData['items'][] = [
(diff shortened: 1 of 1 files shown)


Zales0123:
Hello @jeandanyel! Thank you for that fix. Could you, please, add a spec to the `PayPalItemDataProviderSpec` with data that would previously result in bad decimals and will be good after the change? Thanks in advance! 🚀

liyo (changes requested):
we would need the same update on lines 45 and 50

Zales0123:
Closing in favour of #213 and already a part of 1.2.3 and 1.3.0 🖖 Thank you!

PR #194, opened 2021-04-20
[Bug] Locale support improved

### LINKED ISSUE #188: PayPal plugin KO when using locale "en"
Hi,

I have configured a shop using the locale `en` which didn't give me errors when I installed it, therefore it is a validate locale

then I struggled for hours to understand why the PayPal button was not showing in the shipment method section and here is the thing :  

Your locale  is saved in the cart (table sylius_order)
image

and the plugin PayPal uses this variable in `/vendor/sylius/paypal-plugin/src/Resources/views/payFromCartPage.html.twig` to build the HTTP request to PayPal (for having the javascript library).

When you have `en` (and I presume it does the problem for other 2 chars locales), it automatically adds a suffix `_001`

Check bellow

image

`https://www.paypal.com/sdk/js?client-id=CLIENT_ID_HIDDEN_FOR_GITHUB&locale=en_001&currency=USD"`

This request returns

```
throw new Error("SDK Validation error: 'Invalid query value for locale: en_001'" );

/* Original Error:

Invalid query value for locale: en_001 (debug id: 1497937e2b6bb)

*/
```

Hope it can help some people having this issue!

SirDomin:
fixes #188 

paypal supports only 2 en_XX locales, it changes all en locales to en_US

### DIFF
--- a/src/Processor/LocaleProcessor.php
+++ b/src/Processor/LocaleProcessor.php
@@ -14,6 +14,10 @@ final class LocaleProcessor implements LocaleProcessorInterface
             return $locale;
         }
 
+        if ($locale === 'en') {
+            return 'en_US';
+        }
+
         $locales = array_filter(Locales::getLocales(), function (string $targetLocale) use ($locale): bool {
             return
                 strpos($targetLocale, $locale) === 0 &&
@@ -22,7 +26,7 @@ final class LocaleProcessor implements LocaleProcessorInterface
             ;
         });
 
-        return preg_replace('/en_[A-Z]{2}/', 'en_US', $locales[array_key_first($locales)]);
+        return $locales[array_key_first($locales)];
     }
 
     private function isValidLocale(string $locale): bool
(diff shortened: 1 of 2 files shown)


src/Processor/LocaleProcessor.php
```diff
@@ -17,11 +17,12 @@ public function process(string $locale): string
...
             return
                 strpos($targetLocale, $locale) === 0 &&
-                strpos($targetLocale, '_') !== false
+                strpos($targetLocale, '_') !== false &&
+                strlen($targetLocale) === 5
             ;
         });
 
-        return $locales[array_key_first($locales)];
+        return preg_replace('/en_[A-Z]{2}/', 'en_US', $locales[array_key_first($locales)]);
```
Zales0123, line 25: As this is a special case (I wonder how many more we will notice in the future? 😄), I would do it even before the `array_filter` with a simple:

```php
if ($locale === 'en') {
    return 'en_US';
}
```

And leave this one as it is. Probably the best thing would be to have a table of *default* locales (e.g. `['en' => 'en_US', 'fr' => 'fr_FR', 'it' => 'it_IT']` etc. But that's an idea for the future 🖖

Zales0123 (approved):

Zales0123:
Thanks, @SirDomin! :tada:

bellu:
I have the same issue wit locale **it**.

PR #195, opened 2021-04-21
Build paypal with sylius 1.8 and 1.9

SirDomin:


.github/workflows/build.yml
```diff
@@ -13,20 +13,24 @@ jobs:
...
+        name: "PHP ${{ matrix.php }}, MySQL ${{ matrix.mysql }}, Sylius ${{ matrix.sylius }}"
 
         strategy:
             fail-fast: false
             matrix:
                 php: [7.4, 7.3]
                 node: [10.x]
                 mysql: [5.7, 8.0]
+                symfony: [^4.4, ^5.2]
+                sylius: [~1.8.0, ~1.9.0@beta]
```
Zales0123, line 25: Why do we use 1.9@beta when 1.9 stable is already available?

.github/workflows/build.yml
```diff
@@ -83,6 +87,23 @@ jobs:
                     restore-keys: |
                         ${{ runner.os }}-php-${{ matrix.php }}-composer-
 
+            -
+                name: Install Sylius version
```
Zales0123, line 91: We probably don't need this step as we have `Restrict Sylius version` below

src/Onboarding/Processor/BasicOnboardingProcessor.php
```diff
@@ -76,7 +76,7 @@ public function process(
             'partner_attribution_id' => $response['partner_attribution_id'],
         ]);
 
-        $permissionsGranted = (bool) $request->query->get('permissionsGranted', true);
+        $permissionsGranted = $request->query->get('permissionsGranted') === null ? true : (bool) $request->query->get('permissionsGranted');
```
Zales0123, line 79: This looks strange, the previous implementation should work fine (we have `true` as a default value). Psalm couldn't handle it?

tests/Application/Kernel.php
```diff
@@ -103,4 +99,41 @@ private function isTestEnvironment(): bool
...
+    private function loadRoutesConfiguration(RouteCollectionBuilder $routes, string $confDir): void
+    {
+        $routes->import($confDir . '/{routes}/*' . self::CONFIG_EXTS, '/', 'glob');
+        $routes->import($confDir . '/{routes}/' . $this->environment . '/**/*' . self::CONFIG_EXTS, '/', 'glob');
+        $routes->import($confDir . '/{routes}' . self::CONFIG_EXTS, '/', 'glob');
+    }
+
+    /**
+     * @return BundleInterface[]
+     */
```
Zales0123, line 120: Could be shorter:

```php
/** @return BundleInterface[] */
```

same below

tests/Application/Kernel.php
```diff
@@ -103,4 +99,41 @@ private function isTestEnvironment(): bool
...
+    private function registerBundlesFromFile(string $bundlesFile): iterable
+    {
+        $contents = require $bundlesFile;
+        foreach ($contents as $class => $envs) {
```
Zales0123, line 124: `$envs` => `$environments`? #abbreviationsAreEvil

Zales0123 (changes requested):

src/Onboarding/Processor/BasicOnboardingProcessor.php
SirDomin: in newer versions second parameter must be null || string so thats the way around it :/

composer.json
```diff
@@ -7,7 +7,7 @@
     "require": {
         "php": "^7.3",
 
-        "sylius/sylius": "1.8.*",
+        "sylius/sylius": "~1.8.0 || ~1.9.0@beta",
```
Zales0123, line 10: Isn't `^1.8` enough?

Zales0123 (approved):

PR #206, opened 2021-07-28
Take care about `sylius_core.prepend_doctrine_migrations` parameters

chadyred:
Hello,

It seems that the sylius core boolean parameter "sylius_core.prepend_doctrine_migrations" is not used at all in the prepend function and this cause error into environment which used "sqlite" db, for example, and doesn't want to be aborted with Migrations which deals with a DB connector.

Thanks.


SirDomin:
Thanks, Florian! :1st_place_medal:

chadyred:
Thanks to you for your work

PR #239, opened 2022-04-08
Try to cancel the payment only if it's possible

### LINKED ISSUE #238: [Bug] Transition "cancel" is called even on completed or already canceled payment
**SyliusPayPalPlugin version affected**: >=1.3.2

**Description**  
The "CancelLastPayPalPaymentAction" is called even when the payment page in "unloaded" due to a successful payment. This lead to call the transition "cancel" on the payment even if it's completed and results in an error like this:
```
Uncaught PHP Exception SM\SMException: "Transition "cancel" cannot be applied on state "completed" of object "App\Entity\Payment\Payment" with graph "sylius_payment""
```

**Steps to reproduce**  
1. Place an order with a successful payment
2. You can see in the network tab or in the log (it depends on your monolog settings) that the request to "cancel-last-pay-pal-payment" failed with a 500.

**Possible Solution**  
This is due to [this](https://github.com/Sylius/PayPalPlugin/blob/master/src/Controller/CancelLastPayPalPaymentAction.php#L52). The cancel transition is called without checking if it can be run (with ->can()).
There is also [this point](https://github.com/Sylius/PayPalPlugin/blob/master/src/Controller/CancelPayPalPaymentAction.php#L53) in which this change should be applied.

LucaGallinari:
| Q               | A
| --------------- | -----
| Branch?         | 1.4 (bug fixes, improvements)
| Bug fix?        | yes
| New feature?    | no
| Related tickets | fixes #238

### DIFF
--- a/src/Controller/CancelLastPayPalPaymentAction.php
+++ b/src/Controller/CancelLastPayPalPaymentAction.php
@@ -49,10 +49,12 @@ final class CancelLastPayPalPaymentAction
         $payment = $order->getLastPayment();
 
         $paymentStateMachine = $this->stateMachineFactory->get($payment, PaymentTransitions::GRAPH);
-        $paymentStateMachine->apply(PaymentTransitions::TRANSITION_CANCEL);
+        if ($paymentStateMachine->can(PaymentTransitions::TRANSITION_CANCEL)) {
+            $paymentStateMachine->apply(PaymentTransitions::TRANSITION_CANCEL);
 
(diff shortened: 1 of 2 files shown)


NoResponseMate:
Hi @LucaGallinari :wave:

The fix has been merged into 1.7 via #391; some of the functionality was covered in the meantime, no clue only in one place though 🤔 

Either way, the fix will be available in the next 1.7 release.

Cheers :beers:

PR #255, opened 2022-09-01
fix window reload on javascript payment callbacks

hdasdoria:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | yes
| New feature?    | no
| Related tickets | /

Hi, there was an issue with the javascripts in the payWithPaypal template.

The `window.location.reload()` in error and cancel callbacks was executed instead of being passed as a callback in the `fetch().then()` function.

The page was reloading before the call was made and the payment was never cancelled, so **the payment could not be processed again** !

Had to override locally because half of our paypal orders fell in that and payment were never completed.

### DIFF
--- a/src/Resources/views/payWithPaypal.html.twig
+++ b/src/Resources/views/payWithPaypal.html.twig
@@ -268,7 +268,7 @@
                     method: 'post',
                     headers: {},
                     body: error
-                }).then(window.location.reload());
+                }).then(() => window.location.reload());
             },
             onShippingChange: function(data, actions) {
                 if (!availableCountries.filter(country => country === data.shipping_address.country_code).length) {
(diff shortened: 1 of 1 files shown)


src/Resources/views/payWithPaypal.html.twig
```diff
@@ -268,7 +268,7 @@
                     method: 'post',
                     headers: {},
                     body: error
-                }).then(window.location.reload());
+                }).then(() => window.location.reload());
```
Prometee, line 271:
```suggestion
                }).then(window.location.reload);
```
`window.location.reload` is a function
`window.location.reload()` is the function call

Prometee (changes requested):

PR #297, opened 2024-06-13
Fix and clean up UpdatePayPalOrderAction

maikrosenthal:
| Q               | A
| --------------- | -----
| Branch?         | 1.6 (bug fixes, improvements)
| Bug fix?        | yes
| New feature?    | no

The UpdatePayPalOrderAction fails to update the PayPal order.

The call of ParameterBag::get('shipping_address') should return a scalar value but returns an array.

### DIFF
--- a/src/Resources/config/services/controller.xml
+++ b/src/Resources/config/services/controller.xml
@@ -117,7 +117,6 @@
         <service id="Sylius\PayPalPlugin\Controller\UpdatePayPalOrderAction">
             <argument type="service" id="Sylius\PayPalPlugin\Provider\PaymentProviderInterface" />
             <argument type="service" id="Sylius\PayPalPlugin\Api\CacheAuthorizeClientApiInterface" />
-            <argument type="service" id="Sylius\PayPalPlugin\Api\OrderDetailsApiInterface" />
             <argument type="service" id="Sylius\PayPalPlugin\Api\UpdateOrderApiInterface" />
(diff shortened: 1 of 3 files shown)


src/Resources/config/services/controller.xml
```diff
@@ -117,7 +117,6 @@
         <service id="Sylius\PayPalPlugin\Controller\UpdatePayPalOrderAction">
             <argument type="service" id="Sylius\PayPalPlugin\Provider\PaymentProviderInterface" />
             <argument type="service" id="Sylius\PayPalPlugin\Api\CacheAuthorizeClientApiInterface" />
-            <argument type="service" id="Sylius\PayPalPlugin\Api\OrderDetailsApiInterface" />
```
GSadee, line 120: It is not a BC Break, as it is excluded in our BC Policy, but possibly it would be nice to add a note to the UPGRADE file with such a change

GSadee (approved):

src/Resources/config/services/controller.xml
maikrosenthal: Done

GSadee: Thank you for that, I've moved it to the proper version yet 🖖🏻

GSadee (approved):

GSadee:
Thanks, @maikrosenthal! :tada:

PR #299, opened 2024-06-18
Fix ProcessPayPalOrderAction

maikrosenthal:
| Q               | A
| --------------- | -----
| Branch?         | 1.6 (bug fixes, improvements)
| Bug fix?        | yes
| New feature?    | no

A customer name can contain more than just a first name and last name. It could also contain a salutation, middle names or a title. Using only the first two words of the full_name is therefore not sufficient.

The customer is also added to the address created. This means that the customer can change their shipping address after completing the order and can therefore influence the calculation of shipping costs and tax.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -34,6 +34,22 @@
     )
     ```
 
+   `Sylius\PayPalPlugin\Controller\ProcessPayPalOrderAction`:
+   ```diff
+   public function __construct(
+   -   private readonly OrderRepositoryInterface $orderRepository,
+       private readonly CustomerRepositoryInterface $customerRepository,
+       private readonly FactoryInterface $customerFactory,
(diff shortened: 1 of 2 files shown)


UPGRADE.md
```diff
@@ -71,7 +71,27 @@
    +      private readonly ?RequestFactoryInterface $requestFactory = null,
     )
      ```
-   
+
+   `Sylius\PayPalPlugin\Controller\ProcessPayPalOrderAction`:
```
GSadee: It should be added to the section above `UPGRADE FROM 1.6.0 to 1.6.1`

GSadee (approved):

GSadee (approved):

GSadee:
Thanks, @maikrosenthal! :tada:

PR #331, opened 2025-01-10
Add new addToCart controller

mpysiak:
| Q               | A
| --------------- | -----
| Branch?         | 2.0
| Bug fix?        | yes
| New feature?    | no
| Related tickets | n/a

Fixes bug with (add to cart) live component route returning 404

### DIFF
--- a/src/Controller/AddToCartAction.php
+++ b/src/Controller/AddToCartAction.php
@@ -71,7 +71,7 @@ final readonly class AddToCartAction
             $product = $orderItem->getVariant()->getProduct();
 
             return new RedirectResponse(
-                $this->router->generate('sylius_shop_product_show', ['slug' => $product->getSlug()])
+                $this->router->generate('sylius_shop_product_show', ['slug' => $product->getSlug()]),
             );
         }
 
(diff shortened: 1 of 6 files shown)


GSadee (approved):

src/Controller/AddToCartAction.php
```diff
@@ -0,0 +1,84 @@
...
+        $form = $this->formFactory->create(
+            $formType,
+            $this->addToCartCommandFactory->createWithCartAndCartItem($cart, $orderItem),
+            $configuration->getFormOptions(),
+        );
+
+        $form = $form->handleRequest($request);
+
+        if ($form->isSubmitted() && !$form->isValid()) {
+            return new RedirectResponse((string) $request->headers->get('referer'));
```
NoResponseMate: Having a redirect to a referer in this place seems like a horrible idea. Is this action used in multiple places? Can we not just hardcode a route or maybe split the routing for this controller and use them 1:1 so that the redirect is static?

mpysiak: Referer redirect removed 🙌

GSadee (approved):

PR #364, opened 2025-05-14
Fixes Bug with number rounding

MarkusHolstein:
| Q               | A
| --------------- | -----
| Branch?         | 1.6 (and above)
| Bug fix?        | yes
| New feature?    |no

Und certain circumstances the rounding in PHP runs wild. This leads to a problem, when trying to validate the amount sent to paypal vs. the amount of the cart.

Example:
```
$totalAmount = 0;
$unit['amount']['value'] = '2258.45';

$stringAmount = $unit['amount']['value'] ?? '0';
$totalAmount += (int) ($stringAmount * 100);

var_dump($totalAmount);
var_dump($stringAmount * 100);
```
Results in
```
int(225844)
float(225844.99999999997)
```
Thus, the value needs to be rounded, before casting to int.

So some people are currently unable to use the paypal function of our website.


GSadee:
Hi @MarkusHolstein! Thank you very much for your contribution. I've just merged the fix for this problem in another PR with up-to-date tests for branch 1.7, so I'm closing this one.

PR #374, opened 2025-07-16
Fix problem with repeating PayPal Checkout

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | yes
| Related tickets | 

Proposed solution to the problem of using PayPal Checkout multiple times when there is an uncompleted order.

### DIFF
--- a/src/Controller/CreatePayPalOrderFromCartAction.php
+++ b/src/Controller/CreatePayPalOrderFromCartAction.php
@@ -24,6 +24,8 @@ use Sylius\Component\Core\Payment\Remover\OrderPaymentsRemoverInterface;
 use Sylius\Component\Core\Repository\OrderRepositoryInterface;
 use Sylius\Component\Order\Processor\OrderProcessorInterface;
 use Sylius\PayPalPlugin\DependencyInjection\SyliusPayPalExtension;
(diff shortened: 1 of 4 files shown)


src/Controller/CreatePayPalOrderFromCartAction.php
```diff
@@ -119,20 +130,35 @@ public function __invoke(Request $request): Response
...
         /** @var PaymentInterface $payment */
-        $payment = $order->getLastPayment(PaymentInterface::STATE_CART);
+        $payment = $order->getLastPayment();
+        if ($payment === null) {
+            throw new PaymentNotFoundException();
+        }
+
         /** @var PaymentMethodInterface|null $paymentMethod */
         $paymentMethod = $payment->getMethod();
         $factoryName = $paymentMethod?->getGatewayConfig()?->getFactoryName();
```
NoResponseMate, line 140: This should be moved to ~150 line

NoResponseMate (approved):

GSadee:
I’ve closed this PR because I wasn’t able to reproduce the issue anymore. It’s possible that the problem has been fixed in the meantime.

PR #248, opened 2022-05-24
[MR] Fix Cancel State not working if table sylius_payment too big

sloury:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | yes
| New feature?    | no
| Related tickets | mentioned in #247

PR #392, opened 2025-10-14
Fix performance issue of getting all payments within the payment flow

### LINKED ISSUE #247: Cancel state order not working proprely if table sylius_payment is too big
**SyliusPayPalPlugin version affected**: master

**Description**  
Hi everyone, i had a problem with the cancelation or abort payment step. I have a big database sylius with about more 400000 entries into 'sylius_payment' table. During the cancel process to get all payments, the retriving the id transaction paypal take so much time that cannot work proprely. That redirect me to the payment methods page without the 'cancelled' state or worst get an error 500 'OutofMemory' if you had limited in memory PHP. If i go to the payment methods page, i can't go re-attempt to validate an another transaction.

**Steps to reproduce**  
- Try to get more 10000 entries in 'sylius_payment' table
- Go to the popin Paypal Checkout and click to 'Cancel and return to Test Store' or exit the popin to trigger the cancel state.
- See the time between line 32 - 34 in PaymentProvider.php (too long)

**Possible Solution**  
- Create a new route, like create and complete route, with the orderid sylius
- Pass this new route into PaypalButtonsController.php (line 139) with the orderid sylius
- Modify CancelPaypalPaymentAction line 29 -31 to retrieve the order sylius and the last payment.

NoResponseMate:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| Related tickets | fixes #247, replaces #248

### DIFF
--- a/src/Controller/Webhook/RefundOrderAction.php
+++ b/src/Controller/Webhook/RefundOrderAction.php
@@ -22,6 +22,7 @@ use Sylius\PayPalPlugin\Exception\PaymentNotFoundException;
 use Sylius\PayPalPlugin\Exception\PayPalWrongDataException;
 use Sylius\PayPalPlugin\Provider\PaymentProviderInterface;
 use Sylius\PayPalPlugin\Provider\PayPalRefundDataProviderInterface;
+use Sylius\PayPalPlugin\Repository\Query\PaypalPaymentQueryInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -34,6 +35,7 @@ final class RefundOrderAction
         private readonly PaymentProviderInterface $paymentProvider,
         private readonly ObjectManager $paymentManager,
         private readonly PayPalRefundDataProviderInterface $payPalRefundDataProvider,
+        private readonly ?PaypalPaymentQueryInterface $paypalPaymentQuery = null,
     ) {
         if ($this->stateMachineFactory instanceof FactoryInterface) {
             trigger_deprecation(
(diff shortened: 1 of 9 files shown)


src/Repository/Query/PaypalPaymentQuery.php
```diff
@@ -0,0 +1,114 @@
...
+use Sylius\PayPalPlugin\Exception\PaymentNotFoundException;
+
+final class PaypalPaymentQuery implements PaypalPaymentQueryInterface
+{
+    public function __construct(
+        private readonly EntityManagerInterface $entityManager,
+        private readonly PaymentRepositoryInterface $paymentRepository,
+        private readonly array $updatableStates = ['cart', 'new', 'processing'],
+        private readonly array $cancellableStates = ['cart', 'new', 'processing', 'completed'],
+        private readonly array $refundableStates = ['completed'],
```
GSadee, line 30: I wonder if this should not be taken from the state machine in some way, so that these states remain consistent 🤔

src/Controller/Webhook/RefundOrderAction.php
```diff
@@ -53,7 +55,11 @@ public function __invoke(Request $request): Response
         $refundData = $this->payPalRefundDataProvider->provide($this->getPayPalPaymentUrl($request));
 
         try {
-            $payment = $this->paymentProvider->getByPayPalOrderId((string) $refundData['id']);
+            if (null !== $this->paypalPaymentQuery) {
+                $payment = $this->paypalPaymentQuery->getForRefundingByOrderId((string) $refundData['id']);
+            } else {
+                $payment = $this->paymentProvider->getByPayPalOrderId((string) $refundData['id']);
```
GSadee, line 61: I believe that this service could be deprecated, but preferably in 2.1 🤔

GSadee (approved):

src/Repository/Query/PaypalPaymentQuery.php
NoResponseMate: From what I've searched, there's no way to do that, except by getting the configuration directly (and that's different depending on the used state machine). It might be worth adding something like that directly to the abstraction layer within Sylius.

src/Controller/Webhook/RefundOrderAction.php
NoResponseMate: Added in #398 👋

PR #618, opened 2026-06-22
[BugFix] Fix PayPal Sandbox modal form submitting to wrong URL

Cholin2000:
Sylius AdminBundle's modal-portal.js moves .modal elements to <body> on show.bs.modal. When the .modal lived inside a LiveComponent wrapper div (which carries data-controller="live"), only the modal was relocated, detaching it from its Stimulus ancestor. This broke the data-action="live#action:prevent" binding on the sandbox credentials form, causing a plain HTTP POST to the current page URL instead of an AJAX request to the LiveComponent endpoint, resulting in a MethodNotAllowedHttpException.
| Q               | A
| --------------- | -----
| Branch?         | 2.1 (new features)
| Bug fix?        | yes
| New feature?    | no

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,7 +32,7 @@ jobs:
             matrix:
                 php: ["8.2", "8.3"]
                 symfony: ["^6.4", "^7.4"]
-                sylius: ["~2.0.1", "~2.1.0", "~2.2.0"]
+                sylius: ["~2.1.0", "~2.2.0"]
                 database: ["mysql:8.4", "postgres:16"]
                 node: ["22.x"]
                 state_machine_adapter: ["symfony_workflow"]
@@ -40,8 +40,8 @@ jobs:
                 include:
                     -
                         php: "8.3"
-                        symfony: "~7.3.0"
-                        sylius: "~2.1.0"
+                        symfony: "^7.4"
+                        sylius: "~2.2.0"
                         node: "24.x"
                         database: "mysql:8.4"
                         state_machine_adapter: "winzou_state_machine"
@@ -128,7 +128,7 @@ jobs:
                 uses: actions/upload-artifact@v4
                 if: failure()
                 with:
-                    name: "Behat logs - Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, ${{ matrix.database }}"
+                    name: "Behat logs - Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, ${{ env.DB_TYPE }} ${{ env.DB_VERSION }}"
                     path: etc/build/
                     if-no-files-found: ignore
                     compression-level: 6
--- a/assets/admin/scripts/live-component-modal-portal.js
+++ b/assets/admin/scripts/live-component-modal-portal.js
@@ -1,0 +1,41 @@
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+const SANDBOX_MODAL_COMPONENT_NAME = 'sylius_paypal_sandbox_modal';
+
+function getSandboxModalWrapper(modal) {
+    const parent = modal.parentElement;
+
+    if (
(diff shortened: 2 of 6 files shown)


assets/admin/scripts/live-component-modal-portal.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+document.addEventListener('show.bs.modal', function (event) {
+    const modal = event.target;
+    const parent = modal.parentElement;
+
+    if (parent && parent !== document.body && parent.hasAttribute('data-controller')) {
+        document.body.appendChild(parent);
+        event.stopImmediatePropagation();
+    }
+}, { capture: true });
```
NoResponseMate: This is a global listener; it'll catch all modals with controllers unnecessarily conflicting with other logic. It should only target the sandbox modal.

assets/admin/scripts/live-component-modal-portal.js
```diff
@@ -0,0 +1,37 @@
...
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Fix for Sylius AdminBundle's modal-portal.js breaking Symfony UX LiveComponent
```
NoResponseMate: What is this `modal-portal.js` and where did it come from?

.github/workflows/build.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
                 include:
                     -
                         php: "8.3"
-                        symfony: "~7.3.0"
+                        symfony: "^7.4"
                         sylius: "~2.1.0"
```
NoResponseMate:
```suggestion
                        sylius: "~2.2.0"
```
While we're at it 💃

features/sorting_payment_method_selection.feature
```diff
@@ -12,11 +12,11 @@ Feature: Prioritising PayPal payment method during checkout
         And the store allows paying with "Cash on Delivery" at position 2
         And the store allows paying with "Offline" at position 1
         And the store allows paying with "PayPal" with "PayPal" factory name at position 4
-        And I am a logged in customer
 
     @ui
     Scenario: Seeing payment methods sorted
-        Given I have product "Targaryen T-Shirt" in the cart
+        Given I am a logged in customer
+        And the logged in customer has product "Targaryen T-Shirt" in the cart
```
NoResponseMate: Why? Do we really need a whole new context file and step definition for something that's already defined within the base contexts?

assets/admin/scripts/live-component-modal-portal.js
```diff
@@ -0,0 +1,37 @@
...
+ * This capture-phase listener fires before modal-portal.js (which uses the
+ * bubbling phase). When the modal's direct parent has data-controller, it moves
+ * the entire LiveComponent wrapper to <body> instead of just the modal, keeping
+ * the Stimulus hierarchy intact. event.stopImmediatePropagation() then prevents
+ * modal-portal.js from running and moving just the modal.
+ *
+ * For modals whose parent does NOT have data-controller the listener does nothing,
+ * so modal-portal.js handles them normally.
+ */
+document.addEventListener('show.bs.modal', function (event) {
```
NoResponseMate: Something's broken between this script and the component.
Triggering validation errors within the modal and dismissing it creates web console errors when we want to show the modal again. Refreshing the page fixes it, but it seems something either is not cleaned up properly after dismissal or moved outside of Bootstrap's scope

assets/admin/scripts/live-component-modal-portal.js
Cholin2000: AdminBundle ships it as assets/scripts/modal-portal.js and imports it in its own entrypoint.js. It moves .modal elements to <body> on show.bs.modal to avoid z-index/overflow issues.

assets/admin/scripts/live-component-modal-portal.js
Cholin2000: Fixed by checking parent.getAttribute('data-live-name-value') === 'sylius_paypal_sandbox_modal' instead of the generic hasAttribute('data-controller'), so the listener only targets the PayPal sandbox modal wrapper.

NoResponseMate (approved):

TheMilek (approved):

Issue #674, opened 2026-07-24
Convert XML services configs to PHP


crydotsnake:
Related Issue: https://github.com/Sylius/Sylius/issues/18490

PR #675, opened 2026-07-24
[Maintenance] [XML2PHP] convert service definition configs from xml to PHP

crydotsnake:
| Q               | A
| --------------- | -----
| Branch?         | 2.1
| Bug fix?        | no
| New feature?    | no
| Related tickets | mentioned in #674 

This PR takes care of converting the service definitions from XML to PHP.


tests/TestApplication/config/services_test.php
```diff
@@ -7,7 +7,7 @@
 
     if (str_starts_with($env, 'test')) {
         $container->import('../../../vendor/sylius/sylius/src/Sylius/Behat/Resources/config/services.xml');
```
crydotsnake, line 9: Do we also need to convert the `services.xml` that is referenced here in `sylius/sylius`?

PR #678, opened 2026-08-31
Drop the heavy facilitator

jkindly:


src/Controller/CompleteOnboardingAction.php
```diff
@@ -0,0 +1,94 @@
...
+        /** @var FlashBagInterface $flashBag */
+        $flashBag = $request->getSession()->getBag('flashes');
+        $indexUrl = $this->urlGenerator->generate('sylius_admin_payment_method_index');
+
+        try {
+            $data = (array) json_decode(
+                json: $request->getContent(),
+                associative: true,
+                flags: \JSON_THROW_ON_ERROR,
+            );
```
GSadee, line 55: This endpoint decodes the raw request body as JSON with no Content-Type check and no CSRF token. That makes it vulnerable to the classic "JSON CSRF via text/plain" pattern: a cross-site fetch() with Content-Type: text/plain is a CORS-simple request (no preflight needed), and since we never validate the actual Content-Type here, a JSON-shaped body sent that way still decodes fine.

src/Api/PayPalOnboardingRequestExecutor.php
```diff
@@ -0,0 +1,52 @@
...
+    public function __construct(
+        private ClientInterface $client,
+        private LoggerInterface $logger,
+    ) {
+    }
+
+    public function execute(RequestInterface $request, string $operation): array
+    {
+        try {
+            $response = $this->client->sendRequest($request);
```
GSadee, line 34: `$this->client` here is `sylius.http_client`, which resolves to `Psr\Http\Client\ClientInterface` with no timeout/max_duration override anywhere in this app's framework.yaml. So this call and the same client used by OnboardingTokenApi, SellerCredentialsApi, MerchantOnboardingStatusApi, and PartnerCredentialsProvider, inherits Symfony HttpClient's unbounded defaults instead of failing fast.

Sylius core already has a precedent for exactly this situation: `sylius_admin.http_client` is defined with timeout: 2 / max_duration: 2 specifically for outbound calls made during an admin request.

src/Provider/PartnerCredentialsProvider.php
```diff
@@ -0,0 +1,74 @@
...
+
+        $request = $this->requestFactory->createRequest(Request::METHOD_GET, $this->partnerCredentialsUrl)
+            ->withHeader('Accept', 'application/json')
+        ;
+
+        $content = $this->requestExecutor->execute($request, 'Partner credentials');
+
+        $partnerId = (string) ($content['partner_id'] ?? '');
+        $partnerClientId = (string) ($content['partner_client_id'] ?? '');
+        $partnerLogoUrl = (string) ($content['partner_logo_url'] ?? '');
```
GSadee: Right now a failed/slow call to partner-credentials throws and breaks both the onboarding modal and re-enabling an existing PayPal method. Since `partner_id`/`partner_client_id` are static and identical for every store, we should add a fallback to hardcoded default values when the request fails or times out, instead of hard-failing

config/services/controller.xml
```diff
@@ -84,6 +84,17 @@
             <argument type="service" id="sylius_paypal.enabler.payment_method" />
         </service>
 
+        <service id="sylius_paypal.controller.partner_credentials_test" class="Sylius\PayPalPlugin\Controller\PartnerCredentialsTestAction" />
```
GSadee, line 88: Add `todo` note that this service is to remove

rust-le: fixed

src/Controller/CompleteOnboardingAction.php
rust-le: fixed

src/Api/PayPalOnboardingRequestExecutor.php
rust-le: fixed

src/Provider/PartnerCredentialsProvider.php
rust-le: fixed

PR #680, opened 2026-09-04
Complete the express checkout purchase in the PayPal wallet

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | phase-1
| Bug fix?        | no
| New feature?    | yes
| Related tickets | 

What the plugin calls express was a shortcut into the ordinary checkout: after approving in the wallet the buyer landed on the Sylius summary page and had to press "Place order" a second time before anything was captured. PayPal recorded no express sessions at all.

The approval now captures the payment and completes the order in one request, and the buyer is returned to the thank-you page. The wallet's choices are written back to the order as before, now including the payer's phone number, and the amount is verified against the order total before the order is completed. Both placements ask PayPal for the Pay Now experience and report the page type they render on, which is how a shortcut placement is recognised.

### DIFF
--- a/src/Controller/ProcessPayPalOrderAction.php
+++ b/src/Controller/ProcessPayPalOrderAction.php
@@ -95,6 +95,7 @@ final readonly class ProcessPayPalOrderAction
         $purchaseUnit = (array) $data['purchase_units'][0];
 
         $address = $this->addressFactory->createNew();
+        $address->setPhoneNumber($data['payer']['phone']['phone_number']['national_number'] ?? null);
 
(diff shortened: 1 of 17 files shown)


src/Controller/ProcessPayPalOrderAction.php
NoResponseMate, line 104: Shouldn't we cross-reference this value with the one already present on the payment details?

GSadee: Should be fixed now 👌🏻

NoResponseMate (approved):

PR #681, opened 2026-09-04
Migrate PayPal front-end to Web SDK v6

rust-le:
## Summary

  Migrates the shop-side PayPal button (cart, product, checkout payment page) from JS SDK v5 to the v6 Web SDK

  - `web-sdk/v6/core` loader, one SDK instance per page, `pageType` resolved from context (`checkout`/`cart`/`product-details`)
  - Eligibility via `findEligibleMethods()` instead of `disable-funding` query params
  - Web components (`<paypal-button>`) instead of `paypal.Buttons()`
  - Unified create/capture-order JSON contract across all three placements (`{id, orderId, status, return_url}`)
  - All button JavaScript moved out of Twig `<script>` blocks into a single Stimulus controller, parameterized via `data-*-value` attributes (first Stimulus controller shipped by this plugin — see
  `assets/shop/package.json`)
  - The 6 legacy routes/APIs superseded by this migration are deprecated (not removed), targeting 3.0.0
  - BN code (`PayPal-Partner-Attribution-Id`) now read from a single source everywhere (value itself unchanged, pending Q2)
  - `UPGRADE-2.1.md` documents all BC-relevant changes

  ## Out of scope

  - **Card payments (Hosted Fields → `card-fields` + 3DS)**: SDD v1.5 §4.3.4.2 only documents a vault (save-card) flow, not a plain one-time card payment with 3DS. Implementing this now would mean
  guessing at an unconfirmed API shape — needs a PayPal-confirmed example first.
  - **Final BN code value** (`Sylius_MP_PPCP` per the SDD): blocked on Q2.

 ## Known, pre-existing, unrelated issue found during manual testing

  Cart/product placements can hit a payment-amount mismatch after wallet approval, because the shipping cost estimated before the buyer's address is known can differ from the real cost once the
  address is applied, and PayPal is never informed of the update (its shipping-callback endpoint — `PayPalOrderShippingCallbackAction`, added here — does not appear to be invoked by PayPal for an
  SDK-managed order). Root-caused but not fixed here; unrelated to this PR's scope.


assets/shop/package.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "@sylius/paypal-plugin",
```
GSadee, line 2: For now ok, but we need to decide about the proper naming of plugins assets packages, this one is not consistent with CMS Plugin

src/Creator/PayPalSandboxPaymentMethodCreatorInterface.php
```diff
@@ -19,6 +19,8 @@ interface PayPalSandboxPaymentMethodCreatorInterface
 {
     public const GATEWAY_NAME = 'sylius_paypal_sandbox';
 
+    // TODO: PayPal's SDD specifies "Sylius_MP_PPCP" for this value; not yet confirmed by PayPal
+    // as the correct BN code to send (blocked on an open question to PayPal), so left unchanged.
     public const PARTNER_ATTRIBUTION_ID = 'sylius-ppcp4p-bn-code';
```
GSadee: We should change it to `Sylius_MP_PPCP`

src/Controller/CreatePayPalOrderFromCartAction.php
```diff
@@ -85,7 +85,7 @@ public function __invoke(Request $request): Response
 
         return new JsonResponse([
             'id' => $order->getId(),
-            'orderID' => $payment->getDetails()['paypal_order_id'],
+            'orderId' => $payment->getDetails()['paypal_order_id'],
```
GSadee: We should at least keep the previous ones together with the new keys

GSadee (approved):

src/Provider/PayPalWebSdkConfigurationProvider.php
```diff
@@ -0,0 +1,40 @@
...
+        return sprintf('%s/web-sdk/v6/core', $this->webUrl);
+    }
+
+    public function getInstanceConfig(ChannelInterface $channel, string $pageType): array
+    {
+        return [
+            'clientId' => $this->payPalConfigurationProvider->getClientId($channel),
+            'components' => ['paypal-payments'],
+            'pageType' => $pageType,
+            'partnerAttributionId' => $this->payPalConfigurationProvider->getPartnerAttributionId($channel),
```
NoResponseMate, line 37: IIRC, the `partnerAttributionId` is a static value; the current way of accessing it is horribly convoluted, worth simplifying.

src/Controller/CreatePayPalOrderFromCartAction.php
```diff
@@ -85,6 +85,8 @@ public function __invoke(Request $request): Response
 
         return new JsonResponse([
             'id' => $order->getId(),
+            'orderId' => $payment->getDetails()['paypal_order_id'],
+            // BC with 2.0, where this key was spelled "orderID". Deprecated, removed in 3.0.
```
NoResponseMate:
```suggestion
            // BC with 2.0. Deprecated in 2.1; use "orderId" instead.
```
And put the comment on the same line as the deprecated entry;
Similar below

assets/shop/controllers/PaypalWebSdkController.js
```diff
@@ -0,0 +1,167 @@
...
+        availableCountries: Array,
+        addToCartFormSelector: String,
+        captureOrderUrl: String,
+        cancelOrderUrl: String,
+        errorUrl: String,
+        completeUrl: String,
+        loadingSelector: String,
+    };
+
+    syliusOrderId = null;
```
NoResponseMate, line 21: Have you checked if it's possible to spoof this value through the UI?

src/Provider/PayPalWebSdkConfigurationProvider.php
GSadee: The value isn't static, only the sandbox seed is: production merchants get it from the facilitator's onboarding response into the gateway config, so swapping the DB read for the constant would change the attribution header of every existing shop.

assets/shop/controllers/PaypalWebSdkController.js
GSadee: Good catch, it looks as though there might be a problem here, but it needs to be checked and fixed separately

NoResponseMate (approved):

PR #682, opened 2026-09-08
Show real shipping options in the PayPal shortcut

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | phase-1
| Bug fix?        | no
| New feature?    | yes
| Related tickets | 

The cart and product ("shortcut") placements reach PayPal without a shipping address, so the buyer picks one in the wallet. Until now that address was priced by a client-side handler that wrote a placeholder address (`Temp`/`Temp`/`Temp`) onto the real order and sent PayPal a single, default shipping method — and PayPal's health check reported no shipping-callback endpoint invoked in twelve months.

PayPal now calls a server-side endpoint instead (`POST /pay-pal-order-shipping-callback`), which answers with every shipping method eligible for that address and its price, or a `422` naming the reason the order cannot be shipped there. Nothing is written to the order — the buyer has approved nothing yet. The method they pick is applied to the Sylius order on approval, along with the region, which was previously dropped.

Two resolvers carry the work and are open for decoration: `PayPalShippingOptionsResolverInterface` and `PayPalShippingAddressResolverInterface`. Both build on stock Sylius services, so the wallet offers the same methods and prices the normal checkout does.

### DIFF
--- a/config/routes/shop.yaml
+++ b/config/routes/shop.yaml
@@ -34,6 +34,12 @@ sylius_paypal_shop_update_paypal_order:
     defaults:
         _controller: sylius_paypal.controller.update_paypal_order
 
+sylius_paypal_shop_order_shipping_callback:
+    path: /pay-pal-order-shipping-callback
+    methods: [POST]
+    defaults:
+        _controller: sylius_paypal.controller.paypal_order_shipping_callback
+
 sylius_paypal_shop_create_paypal_order_from_payment_page:
     path: /pay-pal-order-payment-page/{id}/create
     methods: [POST]
--- a/config/routes/shop.yaml
+++ b/config/routes/shop.yaml
@@ -34,12 +34,6 @@ sylius_paypal_shop_update_paypal_order:
     defaults:
         _controller: sylius_paypal.controller.update_paypal_order
 
-sylius_paypal_shop_order_shipping_callback:
-    path: /pay-pal-order-shipping-callback
-    methods: [POST]
-    defaults:
-        _controller: sylius_paypal.controller.paypal_order_shipping_callback
-
 sylius_paypal_shop_create_paypal_order_from_payment_page:
     path: /pay-pal-order-payment-page/{id}/create
     methods: [POST]
--- a/src/Api/CreateOrderApi.php
+++ b/src/Api/CreateOrderApi.php
@@ -13,34 +13,30 @@ declare(strict_types=1);
 
 namespace Sylius\PayPalPlugin\Api;
 
-use Sylius\Bundle\PayumBundle\Model\GatewayConfigInterface;
-use Sylius\Component\Core\Model\AdjustmentInterface;
-use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Core\Model\PaymentInterface;
-use Sylius\Component\Core\Model\PaymentMethodInterface;
 use Sylius\PayPalPlugin\Client\PayPalClientInterface;
+use Sylius\PayPalPlugin\Factory\PayPalOrderFactory;
+use Sylius\PayPalPlugin\Factory\PayPalOrderFactoryInterface;
+use Sylius\PayPalPlugin\Factory\PayPalPurchaseUnitFactory;
 use Sylius\PayPalPlugin\Model\PayPalOrder;
-use Sylius\PayPalPlugin\Model\PayPalPurchaseUnit;
 use Sylius\PayPalPlugin\Provider\PaymentReferenceNumberProviderInterface;
 use Sylius\PayPalPlugin\Provider\PayPalItemDataProviderInterface;
-use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
-use Webmozart\Assert\Assert;
 
 final readonly class CreateOrderApi implements CreateOrderApiInterface
 {
-    public const PAYPAL_INTENT_CAPTURE = 'CAPTURE';
+    public const PAYPAL_INTENT_CAPTURE = PayPalOrder::INTENT_CAPTURE;
 
     public function __construct(
         private PayPalClientInterface $client,
         private PaymentReferenceNumberProviderInterface $paymentReferenceNumberProvider,
         private PayPalItemDataProviderInterface $payPalItemDataProvider,
-        private ?UrlGeneratorInterface $router = null,
+        private ?PayPalOrderFactoryInterface $payPalOrderFactory = null,
     ) {
-        if (null === $this->router) {
+        if (null === $this->payPalOrderFactory) {
             trigger_deprecation(
                 'sylius/paypal-plugin',
                 '2.1',
-                'Not passing $router to "%s" constructor is deprecated and will be prohibited in 3.0',
+                'Not passing $payPalOrderFactory to "%s" constructor is deprecated and will be prohibited in 3.0',
                 self::class,
             );
         }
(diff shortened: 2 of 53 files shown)


src/Payum/Action/CaptureAction.php
```diff
@@ -24,6 +24,14 @@
 
 final readonly class CaptureAction implements ActionInterface
 {
+    /**
+     * PayPal answers CREATED for an order created without a selected payment source, and
+     * PAYER_ACTION_REQUIRED once payment_source.paypal.experience_context is sent. Both mean the order
+     * exists and both carry its id - treating only the former as success leaves the payment without a
+     * paypal_order_id, which breaks every step that follows.
+     */
+    private const ORDER_CREATED_STATUSES = ['CREATED', 'PAYER_ACTION_REQUIRED'];
```
NoResponseMate: Would be nice to extract it to a provider.

src/Api/CreateOrderApi.php
```diff
@@ -74,8 +84,36 @@ public function create(string $token, PaymentInterface $payment, string $referen
...
+    {
+        $callbackUrl = $this->router?->generate(
+            'sylius_paypal_shop_order_shipping_callback',
+            [],
+            UrlGeneratorInterface::ABSOLUTE_URL,
+        );
+
+        if (null === $callbackUrl || !str_starts_with($callbackUrl, 'https://')) {
+            return null;
+        }
```
NoResponseMate: IMO, checking for `https` on a generated route shouldn't be a concern of this service

src/Api/CreateOrderApi.php
```diff
@@ -74,8 +84,36 @@ public function create(string $token, PaymentInterface $payment, string $referen
...
+        );
+
+        $payPalOrder = new PayPalOrder(
+            $order,
+            $payPalPurchaseUnit,
+            self::PAYPAL_INTENT_CAPTURE,
+            $payerReturnUrl,
+            $payerReturnUrl,
+            $this->getShippingCallbackUrl(),
+        );
```
NoResponseMate: How about creating a factory for this?
The `PayPalPurchaseUnit` above could go the same way.

config/routes/shop.yaml
```diff
@@ -34,6 +34,12 @@ sylius_paypal_shop_update_paypal_order:
     defaults:
         _controller: sylius_paypal.controller.update_paypal_order
 
+sylius_paypal_shop_order_shipping_callback:
+    path: /pay-pal-order-shipping-callback
```
NoResponseMate: Completely unrelated; Maybe let's standardize the paths? The urls seem all over the place right now

src/Controller/PayPalOrderShippingCallbackAction.php
```diff
@@ -0,0 +1,173 @@
...
+
+        $total =
+            $this->minorUnits($breakdown, 'item_total') +
+            $this->minorUnits($breakdown, 'tax_total') +
+            $this->minorUnits($breakdown, 'shipping') +
+            $this->minorUnits($breakdown, 'handling') +
+            $this->minorUnits($breakdown, 'insurance') -
+            $this->minorUnits($breakdown, 'discount') -
+            $this->minorUnits($breakdown, 'shipping_discount')
+        ;
```
NoResponseMate: omg

src/Factory/PayPalShippingAddressFactory.php
```diff
@@ -0,0 +1,74 @@
...
+        private RepositoryInterface $provinceRepository,
+    ) {
+    }
+
+    public function resolve(array $payPalShippingAddress): AddressInterface
+    {
+        $countryCode = $this->stringOrNull($payPalShippingAddress['country_code'] ?? null);
+
+        /** @var AddressInterface $address */
+        $address = $this->addressFactory->createNew();
```
NoResponseMate, line 34: This is a mapper/factory, not a resolver 🤔

src/Resolver/PayPalShippingOptionsResolver.php
```diff
@@ -0,0 +1,101 @@
...
+        foreach ($options as $index => $option) {
+            if ((float) $option['amount']['value'] < (float) $options[$cheapest]['amount']['value']) {
+                $cheapest = $index;
+            }
+        }
+
+        $options[$cheapest]['selected'] = true;
+
+        return $options;
+    }
```
NoResponseMate: There's a lot of mapping/transformation logic hidden within various services, most of which are final/readonly or both

src/Controller/ProcessPayPalOrderAction.php
```diff
@@ -148,12 +170,15 @@ public function __invoke(Request $request): Response
...
             $address->setPostcode($purchaseUnit['shipping']['address']['postal_code']);
             $address->setCountryCode($purchaseUnit['shipping']['address']['country_code']);
+            $this->applyProvince($address, (array) $purchaseUnit['shipping']['address']);
 
             $order->setShippingAddress(clone $address);
             $order->setBillingAddress(clone $address);
 
             $this->stateMachineFactory->apply($order, OrderCheckoutTransitions::GRAPH, OrderCheckoutTransitions::TRANSITION_ADDRESS);
 
+            $this->applyShippingMethodSelectedInWallet($order, $purchaseUnit);
```
NoResponseMate, line 180: This change seems untested, and it's quite a big one

src/Factory/PayPalShippingAddressFactory.php
```diff
@@ -0,0 +1,74 @@
...
+            return null;
+        }
+
+        $candidates = null !== $countryCode ? [$countryCode . '-' . $adminArea1, $adminArea1] : [$adminArea1];
+
+        foreach ($candidates as $candidate) {
+            if (null !== $this->provinceRepository->findOneBy(['code' => $candidate])) {
+                return $candidate;
+            }
+        }
```
NoResponseMate, line 59: Guess this mapping logic needs to be included in docs

src/Controller/ProcessPayPalOrderAction.php
GSadee: Covered 👌🏻

src/Factory/PayPalShippingAddressFactory.php
GSadee: Yes and not just this 😓. The documentation as a whole is fairly thin and needs a good number of topics covered properly, not just the address mapping. I'm not adding it in this PR: we're in the middle of heavy development on the next phase, and doing this justice means covering the whole express checkout around it. I've written it down as a separate task to handle before the releases.

One thing in this PR that was simply wrong and is fixed: UPGRADE-2.1.md said the region is matched by
name, but it is matched by code.

NoResponseMate: Yup, but let's note all these things _somewhere_, just so we won't lose some of the context.

src/Resolver/PayPalShippingOptionsResolver.php
```diff
@@ -0,0 +1,115 @@
...
+        foreach ($options as $index => $option) {
+            if ($option->amount() < $options[$cheapest]->amount()) {
+                $cheapest = $index;
+            }
+        }
+
+        $options[$cheapest] = $options[$cheapest]->withSelected(true);
+
+        return $options;
+    }
```
NoResponseMate, line 114: It feels like this logic, returning the selected option, rules for which one is selected by default (guess cheapest atm), could be better encapsulated.
A plain factory that'd return a collection DTO for the options that'd already know which one is selected (and have a specific accessor for it, or whether any is selected), since the rules for it are inside, for instance

NoResponseMate (approved):

src/Resolver/PayPalShippingOptionsResolver.php
GSadee: I will refactor this part in the separate PR 🙌🏻

PR #684, opened 2026-09-09
Add PayLater & Venmo button

rust-le:
Summary

  Adds two v6 Web SDK payment methods on top of the existing PayPal button placements (cart, product, checkout payment page), plus the admin toggles to control them:

  - Pay Later (paypal-pay-later-button): its own payment session (createPayLaterOneTimePaymentSession), gated on findEligibleMethods().isEligible('paylater'), configured with the productCode/countryCode returned by getDetails('paylater').
  - Venmo (venmo-button): its own session (createVenmoOneTimePaymentSession), gated on isEligible('venmo'), with venmo-payments requested on the SDK instance only when enabled.
  - Per-channel admin toggles on the PayPal payment method's gateway config — paylater_enabled, venmo_enabled, messaging_enabled — default on for both new and existing payment methods (missing key treated as true), editable as plain checkboxes.
  - Messaging: toggle + Twig-hook placement plumbing only (div data-paypal-messaging-placeholder). PayPal's own spec for this (SDD §6.1) is a dead link, and the shipped web-sdk/v6/core bundle only lists paypal-messages as a components enum string with no registered custom element for it — no real implementation was
    guessable from what's actually available, so this is left as scaffolding behind the toggle rather than a guessed integration.
  - No backend changes: both new sessions reuse the existing create/capture endpoints and the same createOrder/onApprove/onCancel/onError Stimulus methods the paypal session already uses — the JSON contract is payment-method-agnostic.

### DIFF
--- a/assets/shop/controllers/PaypalWebSdkController.js
+++ b/assets/shop/controllers/PaypalWebSdkController.js
@@ -1,7 +1,7 @@
 import { Controller } from '@hotwired/stimulus';
 
 export default class extends Controller {
-    static targets = ['paypalButton', 'payLaterButton'];
+    static targets = ['paypalButton', 'payLaterButton', 'venmoButton'];
 
     static values = {
         scriptUrl: String,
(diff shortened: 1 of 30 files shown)


rust-le:
Closing; Pay Later and Venmo are now in separate branches.

PR #685, opened 2026-09-09
[SY-833] Enrich paypal order payload

jkindly:
| Q               | A
| --------------- | -----
| Branch?         | 2.1
| Bug fix?        | yes
| New feature?    | no
| Related tickets | SY-833

Today only a total reaches PayPal, which cuts off risk scoring, dispute resolution and package tracking (the Tracking API needs `purchase_units[].items[]`). This enriches the created order payload:

- **Line items** - SKU, category, per-item tax, name, quantity, unit amount, product URL.
- **`amount.breakdown`** that reconciles with the order total, including promotions.
- **Unique `invoice_id`** (survives payment retry) + **`custom_id`** with the Sylius payment id for webhook resolution - single source of truth in `InvoiceNumberProvider`.
- **`experience_context`** - brand name, locale, per-flow shipping/contact preferences, `return_url` == `cancel_url` for app switch.
- **Digital goods** → `NO_SHIPPING` + `DIGITAL_GOODS` (partner-fee orders stay physical).

New constructor args are optional + deprecated (BC-safe), UPGRADE-2.1 updated, tests/PHPStan/ECS green.


src/Model/PayPalOrder.php
```diff
@@ -23,45 +23,57 @@ class PayPalOrder
...
             'intent' => $this->intent,
+            'payment_source' => [
+                'paypal' => [
+                    'experience_context' => $experienceContext,
+                ],
+            ],
             'purchase_units' => [
                 $this->payPalPurchaseUnit->toArray(),
             ],
-            'application_context' => [
```
GSadee, line 54: For now, if I'm correct, removing `application_context` entirely blocks the standard checkout flow with legacy card fields

src/Api/CreateOrderApi.php
```diff
@@ -58,24 +87,65 @@ public function create(string $token, PaymentInterface $payment, string $referen
...
+            $this->provideBrandName($order),
+            $this->provideLocaleCode($order),
+            $paymentPageUrl,
+            $paymentPageUrl,
+        );
 
         return $this->client->post('v2/checkout/orders', $token, $payPalOrder->toArray());
     }
+
+    private function provideBrandName(OrderInterface $order): string
```
GSadee, line 124: If I'm correct, PayPal will reject an empty string, we should probably return `null` here instead, but this needs to be verified

src/Api/CreateOrderApi.php
```diff
@@ -58,24 +87,65 @@ public function create(string $token, PaymentInterface $payment, string $referen
...
+    {
+        /** @var ChannelInterface|null $channel */
+        $channel = $order->getChannel();
+
+        return (string) $channel?->getName();
+    }
+
+    private function provideLocaleCode(OrderInterface $order): string
```
GSadee, line 132: The same as above

src/Api/CreateOrderApi.php
```diff
@@ -58,24 +87,65 @@ public function create(string $token, PaymentInterface $payment, string $referen
...
+        // PayPal expects a BCP 47 locale (e.g. "en-US"), while Sylius stores it as "en_US".
+        return str_replace('_', '-', (string) $order->getLocaleCode());
+    }
+
+    private function providePaymentPageUrl(OrderInterface $order, PaymentInterface $payment): string
+    {
+        Assert::notNull($this->urlGenerator);
+
+        return $this->urlGenerator->generate(
+            'sylius_paypal_shop_pay_with_paypal_form',
```
GSadee, line 143: This route is deprecated, it shouldn't be used

src/Api/CreateOrderApi.php
```diff
@@ -15,25 +15,54 @@
...
     public const PAYPAL_INTENT_CAPTURE = 'CAPTURE';
 
+    private InvoiceNumberProviderInterface $invoiceNumberProvider;
+
     public function __construct(
         private PayPalClientInterface $client,
         private PaymentReferenceNumberProviderInterface $paymentReferenceNumberProvider,
         private PayPalItemDataProviderInterface $payPalItemDataProvider,
+        private ?UrlGeneratorInterface $urlGenerator = null,
+        ?InvoiceNumberProviderInterface $invoiceNumberProvider = null,
```
GSadee, line 43:
```suggestion
        private ?InvoiceNumberProviderInterface $invoiceNumberProvider = null,
```

src/Model/PayPalOrder.php
```diff
@@ -23,45 +23,57 @@ class PayPalOrder
...
 
     public function toArray(): array
     {
+        $experienceContext = array_filter(
+            [
+                'brand_name' => $this->brandName,
+                'locale' => $this->localeCode,
+                'shipping_preference' => $this->getShippingPreference(),
+                'contact_preference' => $this->getContactPreference(),
+                'user_action' => 'PAY_NOW',
```
GSadee, line 49: Why did you remove the constant to use the value directly?

src/Model/PayPalOrder.php
```diff
@@ -23,45 +23,57 @@ class PayPalOrder
...
+                'payment_method_preference' => 'IMMEDIATE_PAYMENT_REQUIRED',
```
GSadee, line 50: I would be for introducing a constant here

src/Api/CreateOrderApi.php
```diff
@@ -58,24 +87,65 @@ public function create(string $token, PaymentInterface $payment, string $referen
...
         return $this->client->post('v2/checkout/orders', $token, $payPalOrder->toArray());
     }
+
+    private function provideBrandName(OrderInterface $order): string
+    {
+        /** @var ChannelInterface|null $channel */
+        $channel = $order->getChannel();
+
+        return (string) $channel?->getName();
+    }
```
GSadee, line 130: After quick check, `brand_name` overrides the business name registered on the merchant's PayPal account. The buyer sees it on the approval sheet, in PayPal's confirmation email and on the receipt. 

Omitting the key looks like the better default: PayPal then falls back to the business name on the merchant's own account.

Because of that, I wonder how to customize these experience context fields. Now, the array is built inside PayPalOrder::toArray() and PayPalOrder is constructed with new in CreateOrderApi::create(), so overriding a single key means decorating CreateOrderApiInterface and reimplementing the whole method. Extracting it would give integrations one small service to decorate: interface ExperienceContextProviderInterface

src/Api/CreateOrderApi.php
```diff
@@ -58,24 +87,65 @@ public function create(string $token, PaymentInterface $payment, string $referen
...
-            $order->getShippingTotal() - $shippingDiscount,
-            (float) $payPalItemData['total_item_value'],
-            (float) $payPalItemData['total_tax'],
-            $order->getOrderPromotionTotal(),
-            (string) $config['merchant_id'],
-            (array) $payPalItemData['items'],
-            $order->isShippingRequired(),
-            $order->getShippingAddress(),
+            referenceId: $referenceId,
+            invoiceNumber: $this->invoiceNumberProvider->provide($payment, $referenceId),
```
GSadee, line 94: I know this field has been sent before, but I’m wondering whether it’s correct 🤔 
PayPal defines this as the external invoice number, it lands in the payer's transaction history, PayPal's emails. Sylius has no invoice at capture time, so we synthesize one instead. Probably we should have a proper extension here, but without sending the invoice number by default 🤔

GSadee (commented):
And the rebase after last merges to phase-1 branch is necessary

src/Model/PayPalOrder.php
```diff
@@ -27,96 +25,52 @@ class PayPalOrder
...
     public function __construct(
-        OrderInterface $order,
-        PayPalPurchaseUnit $payPalPurchaseUnit,
-        string $intent,
-        private readonly ?string $returnUrl = null,
-        private readonly ?string $cancelUrl = null,
-        private readonly ?string $shippingCallbackUrl = null,
+        private readonly PayPalPurchaseUnit $payPalPurchaseUnit,
+        private readonly string $intent,
+        private readonly array $experienceContext = [],
```
GSadee, line 50: I would like us to keep the BC policy as far as possible, so it would be best not to change the existing arguments here, but simply to add new, optional ones that we will use

src/Model/PayPalOrder.php
```diff
@@ -27,96 +25,52 @@ class PayPalOrder
...
                 'paypal' => [
-                    'experience_context' => $this->getExperienceContext($shippingPreference),
+                    'experience_context' => $this->experienceContext,
                 ],
             ];
 
             return $payPalOrder;
         }
 
         $payPalOrder['application_context'] = [
```
GSadee: I wonder if we shouldn't replace `application_context` entirely by `experience_context`

src/Model/PayPalOrder.php
```diff
@@ -27,96 +25,52 @@ class PayPalOrder
...
             'purchase_units' => [
                 $this->payPalPurchaseUnit->toArray(),
             ],
         ];
 
-        // PayPal rejects an order carrying shipping_preference or user_action in both places with
-        // INCOMPATIBLE_PARAMETER_VALUE, so the two context blocks are mutually exclusive.
+        $shippingPreference = $this->experienceContext[self::KEY_SHIPPING_PREFERENCE] ?? self::NO_SHIPPING;
+
         if (self::PAYPAL_ADDRESS === $shippingPreference) {
```
GSadee: We should send `experience_context` basing not on `shippingPreference`, but on payment source. However I believe that it could be fixed after merging #691

src/Provider/ExperienceContextProvider.php
```diff
@@ -0,0 +1,86 @@
...
+final readonly class ExperienceContextProvider implements ExperienceContextProviderInterface
+{
+    public function provide(
+        OrderInterface $order,
+        ?string $returnUrl = null,
+        ?string $cancelUrl = null,
+        ?string $shippingCallbackUrl = null,
+    ): array {
+        // brand_name is intentionally omitted: PayPal then falls back to the business name registered on
+        // the merchant's account. Decorate this provider to send a custom one.
```
GSadee:
```suggestion
```

src/Provider/ExperienceContextProvider.php
```diff
@@ -0,0 +1,86 @@
...
+        $experienceContext = array_filter(
+            [
+                'locale' => $this->provideLocaleCode($order),
+                PayPalOrder::KEY_SHIPPING_PREFERENCE => $this->getShippingPreference($order),
+                'contact_preference' => $this->getContactPreference($order),
+                'user_action' => PayPalOrder::USER_ACTION_PAY_NOW,
+                'payment_method_preference' => PayPalOrder::PAYMENT_METHOD_PREFERENCE_IMMEDIATE,
+                // return_url and cancel_url must be identical for app switch to function.
```
GSadee:
```suggestion
```

src/Provider/PayPalItemDataProvider.php
```diff
@@ -87,7 +118,71 @@ private function addItem(
...
+
+        $itemData['items'][] = $item;
+    }
+
+    /**
+     * PayPal requires the DIGITAL_GOODS category for orders that do not require shipping, so that the
+     * shipping module is skipped. Per the PayPal SDD, orders that use the partner (platform) fee
+     * functionality must stay on PHYSICAL_GOODS regardless of the shipping requirement; this plugin
+     * does not use partner fees, so digital orders are always marked as DIGITAL_GOODS.
+     */
```
GSadee: Remove unnecessary comments

src/Provider/PayPalItemDataProvider.php
```diff
@@ -45,17 +63,26 @@ public function provide(OrderInterface $order): array
...
             $itemValue = $orderItem->getUnitPrice();
 
+            $variant = $orderItem->getVariant();
+            $product = $variant?->getProduct();
+            $itemDetails = [
+                'category' => $category,
+                'sku' => $variant?->getCode(),
+                'description' => $this->resolveDescription($product),
+                'url' => $this->resolveProductUrl($product),
+            ];
```
GSadee: Building this array with additional private methods could be split into separate service

src/Model/PayPalOrder.php
```diff
@@ -27,65 +27,54 @@ class PayPalOrder
...
                 $this->payPalPurchaseUnit->toArray(),
             ],
-        ];
-
-        // PayPal rejects an order carrying shipping_preference or user_action in both places with
-        // INCOMPATIBLE_PARAMETER_VALUE, so the two context blocks are mutually exclusive.
-        if (self::PAYPAL_ADDRESS === $shippingPreference) {
-            $payPalOrder['payment_source'] = [
+            'payment_source' => [
                 'paypal' => [
```
GSadee, line 62: Now, we are sending `paypal` payment source always, even for card payment. We should verify whether this is a right approach

UPGRADE-2.1.md
```diff
@@ -452,7 +453,7 @@
    When no payment awaits payment the endpoint answers `409` instead of raising a `TypeError`, and the
    response carries `orderId` next to the existing `orderID`, with the same value.
 
-1. #### The following signatures changed.
+15. #### The following signatures changed.
```
GSadee, line 456: Not needed changes

UPGRADE-2.1.md
```diff
@@ -496,3 +497,56 @@
...
+18. #### `PayPalPurchaseUnit` and `PayPalOrder` model constructors changed.
+
+   `Sylius\PayPalPlugin\Model\PayPalPurchaseUnit` gained a trailing optional `?string $customId = null`
+   argument; existing positional calls keep working.
+
+   `Sylius\PayPalPlugin\Model\PayPalOrder` no longer takes the order and the individual context fields. It is
+   now constructed with `(PayPalPurchaseUnit $payPalPurchaseUnit, string $intent, array $experienceContext = [])`
+   and its `toArray()` chooses `payment_source.paypal.experience_context` (when the context's
+   `shipping_preference` is `GET_FROM_FILE`) or `application_context` otherwise. Build the `experienceContext`
+   through `ExperienceContextProviderInterface`. If you construct `PayPalOrder` yourself, update the call.
```
GSadee, line 552: It is not true anymore, it should be updated

src/Model/PayPalOrder.php
```diff
@@ -27,65 +27,54 @@ class PayPalOrder
...
+            ],
         ];
-
-        return $payPalOrder;
     }
 
+    /**
+     * @return array<string, mixed>
+     */
     private function getExperienceContext(string $shippingPreference): array
```
GSadee, line 74: Shouldn't we remove this method and use only a specific provider?

src/Provider/PayPalItemDataProvider.php
```diff
@@ -67,33 +83,33 @@ public function provide(OrderInterface $order): array
...
+        return $order->isShippingRequired() ? self::CATEGORY_PHYSICAL_GOODS : self::CATEGORY_DIGITAL_GOODS;
     }
 
-    private function truncateProductName(string $productName): string
+    private function getItemFactory(): PayPalItemFactoryInterface
     {
-        return mb_strlen($productName) > 127
-            ? mb_substr($productName, 0, 124) . '...'
-            : $productName;
+        return $this->itemFactory ?? new PayPalItemFactory();
```
GSadee, line 113: We should define `new PayPalItemFactory()` in the constructor if needed, so as not to create a new instance for every item

GSadee (approved):

PR #686, opened 2026-09-10
Make the order token the capability on the v6 shortcut/payment-page PayPal routes

rust-le:
The cart, payment-page, and process-order endpoints resolved the Sylius order by a raw, sequential id from the path/body with no ownership check, letting anyone who could guess an id overwrite a stranger's addresses and customer, cancel their payment, or read out their PayPal order id. They now resolve by the order's tokenValue instead, GET is dropped from the create-from-cart route, and OrderNotFoundException maps to a real 404 instead of an uncaught 500.


src/Provider/OrderProvider.php
```diff
@@ -47,4 +47,22 @@ public function provideOrderByToken(string $token): OrderInterface
...
+
+    public function provideCartByToken(string $tokenValue): OrderInterface
+    {
+        // Unlike provideOrderByToken()/findOneByTokenValue(), which deliberately excludes orders still in
+        // the "cart" state (it's meant for looking up an already-placed order to pay/re-pay), the v6
+        // shortcut and payment-page placements call this while the order is still being checked out - in
+        // Sylius, an order stays in the "cart" state for the whole checkout process, only flipping to "new"
+        // once it's actually placed. A plain, unfiltered lookup is what "the order I'm currently checking
+        // out, identified by its token" actually means here.
+        /** @var OrderInterface|null $order */
```
GSadee: Not needed comment

src/Provider/OrderProvider.php
```diff
@@ -47,4 +47,22 @@ public function provideOrderByToken(string $token): OrderInterface
 
         return $order;
     }
+
+    public function provideCartByToken(string $tokenValue): OrderInterface
```
GSadee, line 51: The name is misleading as it does not filter by cart

src/Provider/OrderProvider.php
```diff
@@ -47,4 +47,22 @@ public function provideOrderByToken(string $token): OrderInterface
...
+    {
+        // Unlike provideOrderByToken()/findOneByTokenValue(), which deliberately excludes orders still in
+        // the "cart" state (it's meant for looking up an already-placed order to pay/re-pay), the v6
+        // shortcut and payment-page placements call this while the order is still being checked out - in
+        // Sylius, an order stays in the "cart" state for the whole checkout process, only flipping to "new"
+        // once it's actually placed. A plain, unfiltered lookup is what "the order I'm currently checking
+        // out, identified by its token" actually means here.
+        /** @var OrderInterface|null $order */
+        $order = $this->orderRepository->findOneBy(['tokenValue' => $tokenValue]);
```
GSadee, line 66:
```suggestion
        $order = $this->orderRepository->findCartByTokenValue($tokenValue);
```

src/Controller/PayPalButtonsController.php
```diff
@@ -129,6 +145,25 @@ public function renderPaymentPageButtonsAction(Request $request): Response
         }
     }
 
+    // Both button-URL-generation paths below embed the order's tokenValue into the rendered page - but a
+    // cart that reached this page without going through any checkout transition yet (the normal case for
+    // the cart-page placement) never had one assigned (AssignOrderTokenListener only fires on a checkout
+    // *transition*, not on order creation). Assign and persist one now, before generating any URL that
+    // needs it, so the token embedded in this response actually resolves once the buyer clicks the button.
```
GSadee: Do not add not necessary comments in the code

src/Controller/PayPalButtonsController.php
```diff
@@ -129,6 +145,25 @@ public function renderPaymentPageButtonsAction(Request $request): Response
...
+        if (null === $this->orderTokenAssigner || null === $this->orderManager) {
+            throw new \RuntimeException(sprintf(
+                'An instance of "%s" and an order "%s" are required to render the v6 Web SDK placements.',
+                OrderTokenAssignerInterface::class,
+                ObjectManager::class,
+            ));
+        }
+
+        $this->orderTokenAssigner->assignTokenValueIfNotSet($order);
+        $this->orderManager->flush();
```
GSadee: It calls `flush` on every render, on GET calls, we could not resolve it this way. I would be for assigning a token on a POST endpoint

GSadee (commented):
And the rebase is necessary

config/routes/shop.yaml
```diff
@@ -35,13 +35,13 @@ sylius_paypal_shop_update_paypal_order:
...
 sylius_paypal_shop_create_paypal_order_from_payment_page:
-    path: /pay-pal-order-payment-page/{id}/create
+    path: /pay-pal-order-payment-page/{tokenValue}/create
     methods: [POST]
     defaults:
         _controller: sylius_paypal.controller.create_paypal_order_from_payment_page
 
 sylius_paypal_shop_complete_paypal_order_from_payment_page:
-    path: /pay-pal-order-payment-page/{id}/complete
+    path: /pay-pal-order-payment-page/{tokenValue}/complete
```
GSadee, line 64: Changing these routes is a BC break, we should avoid that, provide a new one and deprecate the previous ones, or resolve the initial issue in the different way

rust-le: Rather than a permanently-working deprecated alias (which would keep this IDOR exploitable indefinitely, since deprecation notices go unactioned for a long time), I added an opt-in flag, sylius_paypal.legacy_id_routes_enabled (default false), that restores the old {id}-based routes and body shape when explicitly turned on. This keeps the same "new route, old one still reachable" shape you suggested, but secure by default — see UPGRADE-2.1.md for the full writeup.

src/Provider/OrderProvider.php
rust-le: fixed

src/Provider/OrderProvider.php
rust-le: fixed

src/Controller/PayPalButtonsController.php
rust-le: removed

src/Provider/OrderProvider.php
rust-le: fixed

src/Controller/PayPalButtonsController.php
rust-le: fixed

config/routes/shop.yaml
```diff
@@ -11,8 +11,18 @@ sylius_paypal_shop_create_paypal_order:
         _controller: sylius_paypal.controller.create_paypal_order
 
 sylius_paypal_shop_create_paypal_order_from_cart:
```
GSadee: To maintain the BC policy, we should introduce a new route and not change the previous one, so the name of this one with token should be a new one

config/routes/shop.yaml
```diff
@@ -11,8 +11,18 @@ sylius_paypal_shop_create_paypal_order:
...
+    path: /create-pay-pal-order-from-cart/{tokenValue}
+    methods: [POST]
+    requirements:
+        tokenValue: '(?!\d+(?:/|$)).+'
+    defaults:
+        _controller: sylius_paypal.controller.create_paypal_order_from_cart
+
+sylius_paypal_shop_create_paypal_order_from_cart_legacy:
```
GSadee:
```suggestion
sylius_paypal_shop_create_paypal_order_from_cart:
```

config/routes/shop.yaml
```diff
@@ -11,8 +11,18 @@ sylius_paypal_shop_create_paypal_order:
...
     path: /create-pay-pal-order-from-cart/{id}
-    methods: [POST, GET]
```
GSadee, line 15: Is it intentional that you removed `GET`?

src/Controller/CompletePayPalOrderFromPaymentPageAction.php
```diff
@@ -54,13 +57,19 @@ public function __construct(
...
             );
         }
+        if (null === $this->legacyIdRoutesEnabled) {
+            trigger_deprecation(
+                'sylius/paypal-plugin',
+                '2.1',
+                'Not passing $legacyIdRoutesEnabled to %s constructor is deprecated and will be required in 3.0',
+                self::class,
+            );
+        }
```
GSadee: This is not true to be honest, this flag will disappear in 3.0, the same as legacy routes, so this deprecation notice is redundant

src/Controller/CompletePayPalOrderFromPaymentPageAction.php
```diff
@@ -37,6 +39,7 @@ public function __construct(
         private ObjectManager $orderManager,
         private ?PaymentAmountVerifierInterface $paymentAmountVerifier = null,
         private ?OrderProcessorInterface $orderProcessor = null,
+        private ?bool $legacyIdRoutesEnabled = null,
```
GSadee:
```suggestion
        private ?bool $legacyIdRoutesEnabled = false,
```

src/Controller/CreatePayPalOrderFromCartAction.php
```diff
@@ -38,6 +39,7 @@ public function __construct(
         private ?OrderPaymentsRemoverInterface $orderPaymentsRemover = null,
         private ?OrderProcessorInterface $orderProcessor = null,
         private ?PayPalPaymentMethodsResolverInterface $payPalMethodsResolver = null,
+        private ?bool $legacyIdRoutesEnabled = null,
```
GSadee: The same as above

config/services/listeners/workflow.xml
```diff
@@ -28,5 +28,14 @@
...
         </service>
+
+        <service id="sylius_paypal.listener.workflow.assign_order_token_on_checkout" class="Sylius\PayPalPlugin\EventListener\Workflow\AssignOrderTokenOnCheckoutListener">
+            <argument type="service" id="Sylius\Component\Core\TokenAssigner\OrderTokenAssignerInterface" />
+            <argument type="service" id="sylius.manager.order" />
+
+            <tag name="kernel.event_listener" event="workflow.sylius_order_checkout.completed.address" priority="100" />
+            <tag name="kernel.event_listener" event="workflow.sylius_order_checkout.completed.select_shipping" priority="100" />
+            <tag name="kernel.event_listener" event="workflow.sylius_order_checkout.completed.skip_shipping" priority="100" />
+        </service>
```
GSadee: This listener seems not to be needed as the one above covers the UI path

src/EventListener/Cart/AssignCartTokenListener.php
```diff
@@ -0,0 +1,42 @@
...
+    public function __invoke(GenericEvent $event): void
+    {
+        $command = $event->getSubject();
+        Assert::isInstanceOf($command, AddToCartCommandInterface::class);
+
+        $cart = $command->getCart();
+        Assert::isInstanceOf($cart, OrderInterface::class);
+
+        $this->orderTokenAssigner->assignTokenValueIfNotSet($cart);
+        $this->orderManager->flush();
```
GSadee:
```suggestion
```

assets/shop/controllers/PaypalWebSdkController.js
```diff
@@ -94,7 +94,9 @@ export default class extends Controller {
         }
 
         const data = await response.json();
-        this.syliusOrderId = data.id;
+        if (data.tokenValue) {
```
GSadee: I think it shouldn’t be set again if it’s already been set

assets/shop/controllers/PaypalWebSdkController.js
```diff
@@ -94,7 +94,9 @@ export default class extends Controller {
...
+            this.orderTokenValue = data.tokenValue;
```
GSadee, line 80: Could you verify that it is not possible to overwrite this value by a script? And also the case where we create a PayPal order for a single order and replace its token so that it refers to a different order?

config/routes/shop.yaml
```diff
@@ -11,8 +11,18 @@ sylius_paypal_shop_create_paypal_order:
         _controller: sylius_paypal.controller.create_paypal_order
 
 sylius_paypal_shop_create_paypal_order_from_cart:
+    path: /create-pay-pal-order-from-cart/{tokenValue}
```
GSadee: As we are creating new routes, it would be great to improve the existing paths, to be consistent
```suggestion
    path: /paypal/create-order-from-cart/{tokenValue}
```

config/services/listeners/cart.xml
```diff
@@ -0,0 +1,27 @@
...
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd"
+>
+    <services>
+        <service id="sylius_paypal.listener.cart.assign_cart_token" class="Sylius\PayPalPlugin\EventListener\Cart\AssignCartTokenListener">
+            <argument type="service" id="Sylius\Component\Core\TokenAssigner\OrderTokenAssignerInterface" />
+            <argument type="service" id="sylius.manager.order" />
+
+            <tag name="kernel.event_listener" event="sylius.cart_item_add" priority="100" />
+        </service>
+    </services>
```
GSadee: I would be in favour of not adding this listener either, but instead creating the token as soon as we start creating the PayPal order. We’ve already done this correctly for Express Checkout from the product page; now we need to do the same for the cart and checkout.

src/Controller/CompletePayPalOrderFromPaymentPageAction.php
GSadee: What’s more, it would be a good idea to deprecate this additional argument straight away, as it will be removed in version 3.0

src/DependencyInjection/Configuration.php
```diff
@@ -34,6 +34,7 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootNode
             ->children()
                 ->booleanNode('sandbox')->defaultTrue()->end()
+                ->booleanNode('legacy_id_routes_enabled')->defaultFalse()->end()
```
GSadee: Deprecate this argument as it will be removed in 3.0

assets/shop/controllers/PaypalWebSdkController.js
rust-le: fixed

config/routes/shop.yaml
rust-le: fixed

config/routes/shop.yaml
rust-le: fixed

config/routes/shop.yaml
rust-le: fixed

src/Controller/CompletePayPalOrderFromPaymentPageAction.php
rust-le: fixed

src/Controller/CreatePayPalOrderFromCartAction.php
rust-le: fixed

config/services/listeners/workflow.xml
rust-le: fixed

src/EventListener/Cart/AssignCartTokenListener.php
rust-le: fixed

assets/shop/controllers/PaypalWebSdkController.js
rust-le: fixed

config/services/listeners/cart.xml
rust-le: fixed

src/Controller/CompletePayPalOrderFromPaymentPageAction.php
rust-le: fixed

src/DependencyInjection/Configuration.php
rust-le: fixed

config/routes/shop.yaml
rust-le: Yes, intentional - explained in UPGRADE-2.1.md: GET existed only so AddToCartAction's redirect could reach this state-mutating, PayPal-calling endpoint, which made it reachable from a plain  tag. That's a separate vulnerability from the id-vs-token issue and applies regardless of which route (id- or token-based) is called, so it stays POST-only.

config/routes/shop.yaml
```diff
@@ -10,9 +10,19 @@ sylius_paypal_shop_create_paypal_order:
...
 
+sylius_paypal_shop_create_paypal_order_from_cart_by_token:
+    path: /paypal/create-order-from-cart/{tokenValue}
+    methods: [POST]
+    requirements:
+        tokenValue: '(?!\d+(?:/|$)).+'
+    defaults:
+        _controller: sylius_paypal.controller.create_paypal_order_from_cart
+
 sylius_paypal_shop_create_paypal_order_from_cart:
```
GSadee, line 23: Add comments to old routes that they are deprecated and will be removed in 3.0

src/Controller/PayPalButtonsController.php
```diff
@@ -108,15 +122,16 @@ public function renderPaymentPageButtonsAction(Request $request): Response
         $channel = $this->channelContext->getChannel();
         /** @var OrderInterface $order */
         $order = $this->orderRepository->find($orderId);
+        $this->assignOrderToken($order);
```
GSadee: This controller is designed to render the PayPal button, we cannot make changes to the order, assign a token and call the flush method within it. It shouldn't be a part of this controller.

src/DependencyInjection/Configuration.php
```diff
@@ -34,6 +34,14 @@ public function getConfigTreeBuilder(): TreeBuilder
         $rootNode
             ->children()
                 ->booleanNode('sandbox')->defaultTrue()->end()
+                ->booleanNode('legacy_id_routes_enabled')
+                    ->defaultFalse()
+                    ->setDeprecated(
+                        'sylius/paypal-plugin',
+                        '2.1',
+                        'The "%path%.%node%" option is a temporary migration bridge for the IDOR fix and will be removed in 3.0, along with the legacy, id-based routes it gates.',
```
GSadee:
```suggestion
                        'The "%path%.%node%" option is a temporary migration bridge and will be removed in 3.0, along with the legacy, id-based routes it gates.',
```

assets/shop/controllers/PaypalWebSdkController.js
```diff
@@ -76,7 +76,9 @@ export default class extends Controller {
         }
 
         const data = await response.json();
-        this.syliusOrderId = data.id;
+        if (data.tokenValue && !this.orderTokenValue) {
+            this.orderTokenValue = data.tokenValue;
```
GSadee: Have you checked that there is no possibility of overwriting this value by a script?

src/Controller/PayPalButtonsController.php
GSadee: What I have in mind right now is: drop `{tokenValue}` from the URL generated here and point `createOrderUrl` at a tokenless POST endpoint that resolves the cart from `sylius.context.cart`, assigns the token there, and returns it in the JSON response, the same shape the product page already uses via `sylius_paypal_shop_add_to_cart`. The Stimulus controller picks the token up from that response, exactly as it does today in the express flow, and uses it for the capture call.

UPGRADE-2.1.md
```diff
@@ -214,6 +214,101 @@
...
+1. #### New, token-based routes were added alongside the cart/payment-page/process-order routes — the old ones are unchanged.
+
+   `sylius_paypal_shop_create_paypal_order_from_cart`, `sylius_paypal_shop_create_paypal_order_from_payment_page`,
+   and `sylius_paypal_shop_complete_paypal_order_from_payment_page` take the Sylius order id as a plain,
+   sequential `{id}` path segment; `sylius_paypal_shop_process_paypal_order` takes it as a plain `orderId` in
+   the JSON body. None of the four checked that the caller had any right to that order — an id is guessable,
+   and knowing it was enough to overwrite a stranger's addresses and customer, cancel their payment, or read
+   out their PayPal order id.
+
+   **This is a security fix, not a deprecation, and it does not rename or repurpose any existing route.** Three
```
GSadee:
```suggestion
   Three
```

UPGRADE-2.1.md
```diff
@@ -214,6 +214,101 @@
...
+       legacy_id_routes_enabled: true
+   ```
+
+   This flag is **off by default and reopens the exact IDOR the `_by_token` routes close** — anyone who knows
+   or guesses an order id can act on it again. Only turn it on as a temporary bridge while you migrate callers
+   still depending on the id-based routes, and turn it back off once they are updated. The `{id}`-based routes
+   and `sylius_paypal_shop_process_paypal_order`'s `orderId` body key are always registered/accepted; when the
+   flag is off, they resolve to a `404 Not Found` (or, for `process_paypal_order`, an equivalent rejection)
+   rather than not existing, so the behavior at the HTTP layer is the same either way from a caller's
+   perspective — the flag only decides whether the id is actually honored.
```
GSadee, line 290: The old routes won’t work properly because you’ve changed the JS controller, for example. I’m wondering whether to leave that logic there as a fallback 🤔

config/routes/shop.yaml
rust-le: fixed

rust-le: fixed

src/Controller/PayPalButtonsController.php
rust-le: fixed — exactly as proposed. `PayPalButtonsController` no longer touches the order at all beyond reading it for the render; `CreatePayPalOrderFromCartAction`/`CreatePayPalOrderFromPaymentPageAction` now resolve the current cart from `sylius.context.cart` and assign the token there when no `tokenValue` is given, returning it in the JSON response. `createOrderUrl`/`completeOrderUrl` are generated with no token. Same shape `sylius_paypal_shop_add_to_cart` already used.

src/DependencyInjection/Configuration.php
rust-le: fixed

assets/shop/controllers/PaypalWebSdkController.js
rust-le: Hardened it further: `orderTokenValue` is now a true private class field (`#orderTokenValue`), so it's genuinely inaccessible/unassignable from outside the class, not just a convention. On the second part — a single Stimulus controller instance is scoped to one page load/one order for its whole lifetime, so there's no path for it to be reused across two different orders; a full page reload (the only way to get a new order) also recreates the controller and resets the field.

UPGRADE-2.1.md
rust-le: fixed

UPGRADE-2.1.md
rust-le: Added a paragraph clarifying that the flag only keeps the `{id}`-based HTTP routes reachable for direct/external callers — it doesn't and can't restore the old client-side behavior, since `PaypalWebSdkController.js` and the templates always call the token-based routes now, regardless of the flag.

src/Controller/CreatePayPalOrderFromCartAction.php
```diff
@@ -122,4 +143,39 @@ private function getPayment(OrderInterface $order): PaymentInterface
...
+        if (null === $this->cartContext || null === $this->orderTokenAssigner) {
+            throw new \RuntimeException(sprintf(
+                'An instance of "%s" and "%s" is required to create a PayPal order from the current cart.',
+                CartContextInterface::class,
+                OrderTokenAssignerInterface::class,
+            ));
+        }
+
+        /** @var OrderInterface $cart */
+        $cart = $this->cartContext->getCart();
```
GSadee, line 176: There is a possibility that this cart will be empty, as the last service in the chain creates new cart. We should check that or even better, if there is a payment in proper state on the cart, otherwise return 404.

config/routes/shop.yaml
```diff
@@ -10,9 +10,21 @@ sylius_paypal_shop_create_paypal_order:
     defaults:
         _controller: sylius_paypal.controller.create_paypal_order
 
+sylius_paypal_shop_create_paypal_order_from_cart_by_token:
+    path: /paypal/create-order-from-cart/{tokenValue}
```
GSadee, line 14: You are not using token in the controller, so this route should not have defined token in the path and name, the same applies to some other routes

src/Controller/CompletePayPalOrderFromPaymentPageAction.php
```diff
@@ -118,4 +136,32 @@ private function getTotalPaymentAmountFromPaypal(PaymentInterface $payment): int
...
+        if (!$request->attributes->has('id')) {
+            if (null === $this->cartContext) {
+                throw new \RuntimeException(sprintf(
+                    'An instance of "%s" is required to complete a PayPal order from the current cart.',
+                    CartContextInterface::class,
+                ));
+            }
+
+            /** @var OrderInterface $cart */
+            $cart = $this->cartContext->getCart();
```
GSadee, line 156: In this controller, we should base on token instead of using cart context as this is not a service that initializes PayPal Order

rust-le:
Closing this; the branch introduced a solution that was not approved. Proper olution is in https://github.com/Sylius/PayPalPlugin/pull/701

PR #693, opened 2026-09-15
Drop upgrade-notes renumbering, delegate fallback experience context to the provider, memoize the item factory

jkindly:
| Q               | A
| --------------- | -----
| Branch?         | 2.1 
| Bug fix?        | no
| New feature?    | no


src/Factory/PayPalOrderFactory.php
```diff
@@ -57,6 +57,7 @@ public function create(PaymentInterface $payment, string $referenceId): PayPalOr
             payPalPurchaseUnit: $this->payPalPurchaseUnitFactory->create($payment, $referenceId),
             intent: PayPalOrder::INTENT_CAPTURE,
             experienceContext: $experienceContext,
+            experienceContextProvider: $this->experienceContextProvider,
```
GSadee:
```suggestion
```
Since we pass the `experienceContext` at the same time as creating this `PayPalOrder`, I don’t see the point in passing the `provider` as well. There will be no situation in which we create a `PayPalOrder` without a context, but with a `provider` that we can then use to provide the context. In my view, we should remove the addition of this argument and also leave the removal of the logic for preparing the `experienceContext` from the `PayPalOrder`, as you’ve already done below.

UPGRADE-2.1.md
GSadee: This paragraph is not true anymore

GSadee (approved):

PR #695, opened 2026-09-15
Keep the checkout addresses and their region through express checkout

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | phase-1
| Bug fix?        | yes
| New feature?    | no
| Related tickets | 

Paying with the express PayPal button on an order that was already addressed in the Sylius checkout replaced its shipping and billing addresses with new ones built from PayPal's response. The province was lost (`province_code` and `province_name` came back `NULL`), along with anything else PayPal does not carry and the original rows were left orphaned in `sylius_address`.

What changes:

- Both payloads that carry a shipping address to PayPal now send `admin_area_1`: the province code with the country prefix stripped (`US-TX` → `TX`), or the province name when there is no code. The purchase unit and the pre-capture address patch had to change together, because the patch runs on the express path too and would otherwise strip the region again just before capture.
- `ProcessPayPalOrderAction` no longer rebuilds addresses PayPal could not have changed. An order that reaches the wallet with an address is sent as `SET_PROVIDED_ADDRESS`, which the buyer cannot edit, so it keeps its own addresses. Addresses are still built from PayPal's response when the order carried none.
- The region mapping lives in one place, `PayPalRegionProvider`.


src/Controller/ProcessPayPalOrderAction.php
```diff
@@ -157,23 +157,26 @@ public function __invoke(Request $request): Response
...
+
+                $name = explode(' ', $purchaseUnit['shipping']['name']['full_name']);
+                /** @phpstan-ignore-next-line false positive */
+                $address->setLastName(array_pop($name) ?? '');
+                $address->setFirstName(implode(' ', $name));
+                $address->setStreet($purchaseUnit['shipping']['address']['address_line_1']);
+                $address->setCity($purchaseUnit['shipping']['address']['admin_area_2']);
+                $address->setPostcode($purchaseUnit['shipping']['address']['postal_code']);
+                $address->setCountryCode($purchaseUnit['shipping']['address']['country_code']);
+                $this->applyProvince($address, (array) $purchaseUnit['shipping']['address']);
```
NoResponseMate: Completely aside from the fix, but you could extract address creation here and the `else` below to a separate service 🤷

tests/Unit/Provider/PayPalRegionProviderTest.php
```diff
@@ -0,0 +1,87 @@
...
+        self::assertSame('CA-ON', PayPalRegionProvider::provide($this->address));
+    }
+
+    public function test_it_falls_back_to_the_province_name_when_there_is_no_province_code(): void
+    {
+        $this->address->method('getCountryCode')->willReturn('PL');
+        $this->address->method('getProvinceName')->willReturn('Mazowieckie');
+
+        self::assertSame('Mazowieckie', PayPalRegionProvider::provide($this->address));
+    }
```
NoResponseMate: Data providers can be your friends 🤗

src/Controller/ProcessPayPalOrderAction.php
```diff
@@ -157,23 +157,26 @@ public function __invoke(Request $request): Response
...
-            $address->setFirstName(implode(' ', $name));
-            $address->setStreet($purchaseUnit['shipping']['address']['address_line_1']);
-            $address->setCity($purchaseUnit['shipping']['address']['admin_area_2']);
-            $address->setPostcode($purchaseUnit['shipping']['address']['postal_code']);
-            $address->setCountryCode($purchaseUnit['shipping']['address']['country_code']);
-            $this->applyProvince($address, (array) $purchaseUnit['shipping']['address']);
-
-            $order->setShippingAddress(clone $address);
-            $order->setBillingAddress(clone $address);
+            if (null === $order->getShippingAddress()) {
```
NoResponseMate, line 168: A rather general question: can we change the address on PayPal's side?
And if so, is that case covered/how should it be handled?

GSadee: Currently, if we start the Express Checkout from the product page or the cart without having completed the addressing step, then yes, we can select an address on the PayPal page. Otherwise, if we go there, for example, from the payment step page, where we have already completed the addressing step, then we cannot change the address

src/Controller/ProcessPayPalOrderAction.php
GSadee: I was thinking the same thing yesterday, but I didn’t want to do too much refactoring here, as it was already in place

TheMilek (approved):

PR #699, opened 2026-09-16
SY-838: Send package tracking to PayPal

jkindly:
| Q               | A
| --------------- | -----
| Branch?         | phase-1
| Bug fix?        | no
| New feature?    |yes
| Related tickets | 

When order is paid via PayPal, the new form is shown in Shipments section. After submit, tracking code will be sent to PayPal and will be visible in customer order in PayPal dashboard.

image
image
image
image


src/DependencyInjection/Configuration.php
```diff
@@ -26,6 +26,19 @@ final class Configuration implements ConfigurationInterface
...
+     */
+    private const DEFAULT_TRACKING_CARRIERS = [
+        'FEDEX', 'UPS', 'USPS', 'DHL', 'DHL_EXPRESS', 'DHL_GLOBAL_MAIL', 'DPD', 'DPD_LOCAL',
+        'GLS', 'TNT', 'ARAMEX', 'AMAZON_MCF', 'DEUTSCHE_POST', 'DHL_DEUTSCHE_POST', 'HERMES',
+        'ROYAL_MAIL', 'PARCELFORCE', 'COLISSIMO', 'CHRONOPOST_FRANCE', 'POCZTA_POLSKA',
+        'INPOST_PACZKOMATY', 'DPD_POLAND', 'CORREOS_SPAIN', 'POSTNL', 'BPOST',
+        'POSTNORD_SVERIGE', 'POSTEN_NORGE', 'AUSTRIAN_POST', 'SWISS_POST', 'POSTE_ITALIANE',
+        'CANADA_POST', 'PUROLATOR', 'AUSTRALIA_POST', 'JAPAN_POST', 'CHINA_POST', 'SF_EXPRESS',
+        'INDIA_POST',
+    ];
```
GSadee, line 40: This list isn’t complete. PayPal lists many more carriers in its documentation, can we retrieve them via the API? I can see below that we can define our own list, which is fine, but by default we could retrieve them all. Wdyt? 🤔

src/Entity/ShipmentTracking.php
```diff
@@ -0,0 +1,158 @@
...
+    private ?string $carrier = null;
+
+    #[ORM\Column(name: 'carrier_name_other', type: 'string', nullable: true)]
+    private ?string $carrierNameOther = null;
+
+    #[ORM\Column(name: 'paypal_tracker_id', type: 'string', nullable: true)]
+    private ?string $payPalTrackerId = null;
+
+    #[ORM\Column(name: 'state', type: 'string', length: 32)]
+    private string $state = self::STATE_PENDING;
```
GSadee, line 43: I’m wondering whether this name is a bit misleading, because it’s not actually the shipment status, but the tracking synchronisation status and whether we’ve sent it to PayPal, if I’ve understood correctly. But maybe I’m wrong, and that’s fine. Wdyt? 🤔

src/Manager/ShipmentTrackingManager.php
```diff
@@ -0,0 +1,48 @@
...
+        private EntityManagerInterface $entityManager,
+    ) {
+    }
+
+    public function updateCarrier(ShipmentInterface $shipment, ?string $carrier, ?string $carrierNameOther): void
+    {
+        $tracking = $this->shipmentTrackingRepository->findOneByShipment($shipment);
+
+        if (null === $tracking) {
+            $tracking = new ShipmentTracking($shipment);
```
GSadee, line 36: Use a factory instead of `new` directly here

src/Migrations/Version20260909120000.php
```diff
@@ -0,0 +1,54 @@
...
+        $table->addColumn('updated_at', 'datetime', ['notnull' => false]);
+        $table->setPrimaryKey(['id']);
+        $table->addUniqueIndex(['shipment_id'], 'UNIQ_paypal_shipment_tracking_shipment');
+        $table->addForeignKeyConstraint(
+            'sylius_shipment',
+            ['shipment_id'],
+            ['id'],
+            ['onDelete' => 'CASCADE'],
+            'FK_paypal_shipment_tracking_shipment',
+        );
```
GSadee, line 47: Use `$this->addSql('CREATE TABLE...)...`. The same below

src/Provider/OrderPayPalPaymentProvider.php
```diff
@@ -0,0 +1,61 @@
...
+            if (SyliusPayPalExtension::PAYPAL_FACTORY_NAME !== $gatewayConfig->getFactoryName()) {
+                continue;
+            }
+
+            $details = $payment->getDetails();
+            if (!isset($details['paypal_order_id'])) {
+                continue;
+            }
+
+            $payPalPayment = $payment;
```
GSadee, line 56: Shouldn't be returned here?

src/Entity/ShipmentTracking.php
```diff
@@ -0,0 +1,158 @@
...
+
+namespace Sylius\PayPalPlugin\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Sylius\Component\Core\Model\ShipmentInterface;
+use Sylius\PayPalPlugin\Repository\ShipmentTrackingRepository;
+
+#[ORM\Entity(repositoryClass: ShipmentTrackingRepository::class)]
+#[ORM\Table(name: 'sylius_paypal_plugin_shipment_tracking')]
+class ShipmentTracking implements ShipmentTrackingInterface
```
GSadee, line 22: I’m wondering whether, for the sake of consistency with Sylius, we shouldn’t make this a resource, although I’m not sure if that makes complete sense, wdyt?

src/Form/Extension/ShipmentShipTypeExtension.php
```diff
@@ -0,0 +1,165 @@
...
+            $form->get('carrier')->addError(new FormError('sylius_paypal.shipment_tracking.carrier_required'));
+
+            return;
+        }
+
+        if (null !== $carrier && null === $carrierNameOther && $this->carrierProvider->isOther($carrier)) {
+            $form->get('carrier_name_other')->addError(new FormError('sylius_paypal.shipment_tracking.carrier_name_other_required'));
+
+            return;
+        }
```
GSadee, line 122: It should be handled by the validator and there should be validation messages, not 500 exceptions

src/Processor/ShipmentTrackingProcessor.php
GSadee, line 1: I'm not sure if the flushes here are correct, could you verify that? You can consider `DispatchAfterCurrentBusStamp` here

src/Entity/ShipmentTracking.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        return $this->updatedAt;
+    }
+
+    public function markAsSynced(?string $payPalTrackerId): void
+    {
+        $this->state = self::STATE_SYNCED;
+        $this->payPalTrackerId = $payPalTrackerId;
+        $this->lastError = null;
+        ++$this->attempts;
```
GSadee, line 136: Not sure if it should also increment attempts 🤔

src/Manager/ShipmentTrackingManager.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Sylius\PayPalPlugin\Manager;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Sylius\Component\Core\Model\ShipmentInterface;
+use Sylius\PayPalPlugin\Entity\ShipmentTracking;
+use Sylius\PayPalPlugin\Provider\CarrierProviderInterface;
+use Sylius\PayPalPlugin\Repository\ShipmentTrackingRepositoryInterface;
+
+final readonly class ShipmentTrackingManager implements ShipmentTrackingManagerInterface
```
GSadee, line 22: There is no test class for this one

src/Message/Handler/SendShipmentTrackingHandler.php
```diff
@@ -0,0 +1,39 @@
...
+declare(strict_types=1);
+
+namespace Sylius\PayPalPlugin\Message\Handler;
+
+use Sylius\Component\Core\Model\ShipmentInterface;
+use Sylius\PayPalPlugin\Message\SendShipmentTracking;
+use Sylius\PayPalPlugin\Processor\ShipmentTrackingProcessorInterface;
+use Sylius\Resource\Doctrine\Persistence\RepositoryInterface;
+
+final readonly class SendShipmentTrackingHandler
```
GSadee, line 21: There is no test class for this one

GSadee (commented):
As this is a quite large functionality with a few new services dependent on it, I’m wondering whether it might be better to move it to a separate namespace called `PackageTracking`

jkindly:
@GSadee 
The rest of the plugin is organised by type rather than by feature - Api/, Provider/, Processor/, Factory/, Manager/ - and the tests mirror that layout one-to-one. Moving just this feature into PackageTracking\* would make it the only feature-first namespace here, so a reader would have to know which convention applies before looking for a class.

src/DependencyInjection/Configuration.php
jkindly: There's no endpoint for it - PayPal publishes the carrier list only as a static table in the docs. So "retrieving them all" would mean vendoring that enum, not calling the API.
I left the curated default in place, because 810 entries in a plain `<select>` is unusable and every code would also need a translation key to avoid showing raw `ABXEXPRESS_MY`-style labels. The default is only a default: `sylius_paypal.tracking.carriers` replaces it entirely, `OTHER` + carrier_name_other covers anything missing, and an unknown code falls back to rendering the code itself.
If we do want full coverage, I'd suggest a follow-up: ship the generated 810-code list and swap the select for an autocomplete so the dropdown stays usable.

src/PackageTracking/Entity/ShipmentTracking.php
```diff
@@ -0,0 +1,158 @@
...
+    private ?string $carrier = null;
+
+    #[ORM\Column(name: 'carrier_name_other', type: 'string', nullable: true)]
+    private ?string $carrierNameOther = null;
+
+    #[ORM\Column(name: 'paypal_tracker_id', type: 'string', nullable: true)]
+    private ?string $payPalTrackerId = null;
+
+    #[ORM\Column(name: 'state', type: 'string', length: 32)]
+    private string $state = self::STATE_PENDING;
```
jkindly: You read it right - it's the state of the tracking record's synchronisation with PayPal, not the shipment's state. I kept state because it's the Sylius convention for a state field on an entity, and the entity is ShipmentTracking, so `ShipmentTracking::getState()` reads as "state of the tracking", not "state of the shipment". The values (pending / synced / failed) and the admin label (sylius_paypal.shipment_tracking.paypal_sync) say explicitly it's about the PayPal sync. If you'd still prefer it explicit in the field name I'll rename it to `syncState`.

src/PackageTracking/Entity/ShipmentTracking.php
```diff
@@ -0,0 +1,158 @@
...
+
+namespace Sylius\PayPalPlugin\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Sylius\Component\Core\Model\ShipmentInterface;
+use Sylius\PayPalPlugin\Repository\ShipmentTrackingRepository;
+
+#[ORM\Entity(repositoryClass: ShipmentTrackingRepository::class)]
+#[ORM\Table(name: 'sylius_paypal_plugin_shipment_tracking')]
+class ShipmentTracking implements ShipmentTrackingInterface
```
jkindly: I'd keep it a plain entity for now. It has no admin CRUD, no grid, no API resource - registering it as a Sylius resource would add the unused services.

src/PackageTracking/Entity/ShipmentTracking.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        return $this->updatedAt;
+    }
+
+    public function markAsSynced(?string $payPalTrackerId): void
+    {
+        $this->state = self::STATE_SYNCED;
+        $this->payPalTrackerId = $payPalTrackerId;
+        $this->lastError = null;
+        ++$this->attempts;
```
jkindly: Deliberate - attempts counts calls made to PayPal, not failures, so a successful first send shows "Attempts: 1" rather than 0, and the admin can tell "synced on the first try" from "synced after three retries". lastError/state already carry the failure information, so a separate failure counter would be redundant.

src/Provider/OrderPayPalPaymentProvider.php
jkindly: The accumulator was the second half of a sandbox bugfix (PayPal order status "CREATED" is not eligible for tracking - an abandoned earlier attempt shadowed the completed payment), so first-match would regress it; the plugin uses `getLastPayment()` everywhere for the same reason. Kept last-wins and got the early return by iterating in reverse, plus a test that actually pins it.

src/DependencyInjection/Configuration.php
GSadee: OK, let’s leave it as it is for now

src/Processor/ShipmentTrackingProcessor.php
```diff
@@ -0,0 +1,210 @@
...
+            'items' => $this->itemsProvider->provide($shipment),
+        ];
+
+        if ($this->carrierProvider->isOther($carrier)) {
+            $body['carrier_name_other'] = (string) $tracking->getCarrierNameOther();
+        }
+
+        $response = $this->addTrackingApi->add($token, $payPalOrderId, $body);
+
+        if (isset($response['name']) || isset($response['debug_id'])) {
```
GSadee: You are checking only name and debug_id, so if get for example `401 {"error":"invalid_token",…}` or 500 in the response, even with empty body, the tracking will be marked as synced

src/PackageTracking/Processor/ShipmentTrackingProcessor.php
```diff
@@ -0,0 +1,210 @@
...
+        $token = $this->authorizeClientApi->authorize($paymentMethod);
+
+        $orderDetails = $this->orderDetailsApi->get($token, $payPalOrderId);
+
+        if (!isset($orderDetails['status'])) {
+            throw new PayPalApiErrorException(sprintf('GET v2/checkout/orders/%s', $payPalOrderId), $orderDetails);
+        }
+
+        $status = (string) $orderDetails['status'];
+        if (!in_array($status, self::ELIGIBLE_ORDER_STATUSES, true)) {
```
GSadee, line 132: According to SDD §7.1, in addition to the status, we should also check whether `$orderDetails[“purchase_units”][0][“items”]` have been dispatched

src/PackageTracking/Console/Command/SendShipmentTrackingCommand.php
GSadee, line 1: It would be good to process these shipments in batches or at least for now, add a limit option to the command

GSadee:
> @GSadee The rest of the plugin is organised by type rather than by feature - Api/, Provider/, Processor/, Factory/, Manager/ - and the tests mirror that layout one-to-one. Moving just this feature into PackageTracking* would make it the only feature-first namespace here, so a reader would have to know which convention applies before looking for a class.

Fair point, the plugin is type-first and the tests mirror it, so mixing conventions isn't free.

Two things still push me the other way. This feature adds 34 classes across 19 directories and modifies exactly two existing files (both DI wiring), I don't think we've had anything this self-contained here in a long time. And the type-first convention isn't really preserved by leaving it: the PR creates src/Dispatcher/, src/Message/ and src/Validator/Constraints/ as new top-level buckets, each holding package-tracking classes and nothing else.

The rule stays easy to state: a feature namespace only for something genuinely self-contained, type namespaces for anything shared, PayPalApiErrorException is thrown for the order-details call too, so it stays in src/Exception/.

Longer term I'd like us to consider the same for other features around the next major. If we never take the first one, the answer is always "it would be the only one". Since the tests mirror the layout, the move is mechanical.

So I'm in favour of doing it now. No strong opinion on the name: `PackageTracking` is precise, `Shipment` might age better if we add anything else on that side. What do you think?

src/Processor/ShipmentTrackingProcessor.php
jkindly: I considered checking the HTTP status code, but PayPalClientInterface only returns the decoded body, so exposing the status would mean a BC break on the client interface and touching every API class in the plugin, which is too much for this PR. Instead I flipped the check to a positive one: the tracking call only counts as successful when PayPal returns the order (id present, per the order schema in the Orders v2 spec), so OAuth errors, empty 5xx bodies and any other error response now throw and mark the record as failed.

GSadee (approved):

PR #702, opened 2026-09-17
[WIP] Add Venmo as a v6 payment method (SY-835)

rust-le:


src/Provider/PayPalConfigurationProvider.php
```diff
@@ -48,6 +48,11 @@ public function isPayLaterEnabled(ChannelInterface $channel): bool
         return (bool) ($this->getPayPalPaymentMethodConfig($channel)['pay_later_enabled'] ?? true);
     }
 
+    public function isVenmoEnabled(ChannelInterface $channel): bool
+    {
+        return (bool) ($this->getPayPalPaymentMethodConfig($channel)['venmo_enabled'] ?? true);
```
GSadee: I believe it should be false by default

src/Controller/PayPalButtonsController.php
```diff
@@ -156,16 +158,29 @@ public function renderPaymentPageButtonsAction(Request $request): Response
...
         }
     }
 
+    /** @return array<int, string> */
+    private function getWebSdkComponents(ChannelInterface $channel): array
+    {
+        $components = PayPalWebSdkConfigurationProviderInterface::DEFAULT_COMPONENTS;
+
+        if ($this->getFundingSourcesConfigurationProvider()->isVenmoEnabled($channel)) {
+            $components[] = 'venmo-payments';
```
GSadee: Add `venmo-payments` as const

GSadee (commented):
- There is no Venmo in `PayPalPaymentSourceProvider` service, so Venmo payments are assigned the default payment source set up for PayPal, rather than for Venmo, which is what they expect
- There is no note added to the upgrade file with the changes

src/Provider/PayPalConfigurationProvider.php
rust-le: fixed

src/Controller/PayPalButtonsController.php
rust-le: fixed

config/app/twig_hooks/admin/payment_method/update.yaml
```diff
@@ -21,6 +21,9 @@ sylius_twig_hooks:
             pay_later_enabled:
                 template: '@SyliusPayPalPlugin/admin/payment_method/form/section/gateway_configuration/paylater_enabled.html.twig'
                 priority: -500
+            venmo_enabled:
+                template: '@SyliusPayPalPlugin/admin/payment_method/form/section/gateway_configuration/venmo_enabled.html.twig'
+                priority: -600
```
GSadee: The messaging component is directly connected with the PayLater, so I believe it should be right after PayLater and Venmo should then be listed below them. The same applies to all pages in the shop (judging by the screenshots) where the buttons are displayed.

config/app/twig_hooks/admin/payment_method/update.yaml
```diff
@@ -21,6 +21,9 @@ sylius_twig_hooks:
...
             messaging_enabled:
                 template: '@SyliusPayPalPlugin/admin/payment_method/form/section/gateway_configuration/messaging_enabled.html.twig'
                 priority: -600
```
GSadee: You've added two hooks with the same priority, adjust it

src/Controller/PayPalButtonsController.php
```diff
@@ -156,16 +160,29 @@ public function renderPaymentPageButtonsAction(Request $request): Response
...
             return new Response('');
         }
     }
 
+    /** @return array<int, string> */
+    private function getWebSdkComponents(ChannelInterface $channel): array
+    {
+        $components = PayPalWebSdkConfigurationProviderInterface::DEFAULT_COMPONENTS;
+
+        if ($this->getFundingSourcesConfigurationProvider()->isVenmoEnabled($channel)) {
```
GSadee: For every render, there are two queries to see if Venmo is enabled, one here and one when constructing the array for the `venmoEnabled` variable. Try to optimise this.

src/Form/Type/PayPalConfigurationType.php
```diff
@@ -45,6 +45,7 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ->add('reports_sftp_username', TextType::class, ['label' => 'sylius_paypal.sftp_username', 'required' => false])
             ->add('reports_sftp_password', TextType::class, ['label' => 'sylius_paypal.sftp_password', 'required' => false])
             ->add('pay_later_enabled', CheckboxType::class, ['label' => 'sylius_paypal.paylater_enabled', 'required' => false])
+            ->add('venmo_enabled', CheckboxType::class, ['label' => 'sylius_paypal.venmo_enabled', 'required' => false])
             ->add('messaging_enabled', CheckboxType::class, ['label' => 'sylius_paypal.messaging_enabled', 'required' => false])
```
GSadee:
```suggestion
            ->add('messaging_enabled', CheckboxType::class, ['label' => 'sylius_paypal.messaging_enabled', 'required' => false])
            ->add('venmo_enabled', CheckboxType::class, ['label' => 'sylius_paypal.venmo_enabled', 'required' => false])
```

src/Provider/PayPalPaymentSourceProvider.php
```diff
@@ -26,12 +26,13 @@ public function provide(OrderInterface $order, string $paymentSource, array $exp
             self::GOOGLE_PAY => [self::GOOGLE_PAY => [
                 'attributes' => ['verification' => ['method' => PayPalOrder::VERIFICATION_METHOD_SCA_WHEN_REQUIRED]],
             ]],
+            self::VENMO => [self::VENMO => ['experience_context' => $experienceContext]],
```
GSadee: The current `$experienceContext` you are using is the PayPal context: `locale, contact_preference, user_action, payment_method_preference, return_url, cancel_url, app_switch_preference`. If I checked correctly, the PayPal `venmo_wallet_experience_context` only accepts `brand_name`, `shipping_preference` and `order_update_callback_config`. 
Each source has its own structure, and it appears that we are passing an incorrect experience context here.

assets/shop/controllers/PaypalWebSdkController.js
```diff
@@ -93,6 +94,10 @@ export default class extends Controller {
                 this.payLaterButtonTarget.countryCode = payLaterDetails.countryCode;
                 this.wireUpButton(this.payLaterButtonTarget, this.sdkInstance.createPayLaterOneTimePaymentSession(this.buildSessionOptions()));
             }
+
+            if (this.venmoEnabledValue && this.hasVenmoButtonTarget && paymentMethods.isEligible('venmo')) {
```
GSadee: There is no guard `!this.wiredTargets.has` for venmo as it is for paypal or paylater above

config/app/twig_hooks/admin/payment_method/update.yaml
rust-le: fixed, proper order now

config/app/twig_hooks/admin/payment_method/update.yaml
rust-le: fixed

src/Controller/PayPalButtonsController.php
rust-le: Fixed in all three render methods, not just this one — isVenmoEnabled() is now called once into a local $venmoEnabled, reused for both getWebSdkComponents() and the template variable.

src/Form/Type/PayPalConfigurationType.php
rust-le: Applied

src/Provider/PayPalPaymentSourceProvider.php
rust-le: Fixed

assets/shop/controllers/PaypalWebSdkController.js
rust-le: Added the same !this.wiredTargets.has('venmo') / .add('venmo') guard

PR #704, opened 2026-09-18
Add Pay Later to the regular (post-checkout) PayPal payment page

rust-le:
| Q               | A
| --------------- | -----
| Branch?         | phase-1
| Bug fix?        | no
| New feature?    | no
| Related tickets | SY-858

Pay Later was already available on the three "express" placements (product page, cart, and the
checkout/select-payment wallet button) but missing from the classic `pay_with_paypal` page —
the one shown when PayPal is picked as a plain payment method rather than through an express
shortcut. This adds the Pay Later button and eligibility messaging there too, reusing the same
Web SDK session/messaging pattern already proven on the other three placements.

Firefox_Screenshot_2026-09-18T09-50-36 260Z


assets/shop/scripts/paypal-payment-page.js
GSadee, line 1: Session is memoized module-wide, so the first controller to connect fixes the eligibility request for the whole page and PaymentPageCardFieldsController, and PaymentPageGooglePayController both call paymentPageSession() without amount. All three are lazy chunks, so the winner is whichever one loads first; when it isn't this one, findEligibleMethods() runs with no amount and Pay Later eligibility + productCode are resolved against nothing.

Could you verify that and probably pass amount from the other two as well?

templates/shop/pay_with_paypal/content/paypal_messaging.html.twig
```diff
@@ -0,0 +1,13 @@
+{% if sylius_paypal_is_messaging_enabled() %}
+    {% set order = hookable_metadata.context.order %}
+
+    <paypal-message
+        class="d-block mb-3"
+        {{ stimulus_controller('@sylius/paypal-plugin/paypal-message', {
+            amount: (order.total / 100)|number_format(2, '.', ''),
```
GSadee: There is inconsistency with the amount source for the messaging, once it’s taken from the order, and once from the payment, it should be consistent and the source should be payment I believe 🤔

templates/shop/pay_with_paypal/content/paypal_messaging.html.twig
```diff
@@ -0,0 +1,13 @@
...
+            currencyCode: order.currencyCode,
```
GSadee:
```suggestion
            amount: context.amount
            currencyCode: context.currencyCode,
```
🤔

assets/shop/controllers/PaymentPageWalletButtonController.js
```diff
@@ -20,28 +22,42 @@ export default class extends Controller {
...
                 currencyCode: this.currencyCodeValue,
+                amount: this.hasAmountValue ? this.amountValue : undefined,
                 createOrderUrl: this.createOrderUrlValue,
             });
 
-            if (!session.isEligible('paypal')) {
-                return;
+            this.session = session;
+
+            if (session.isEligible('paypal')) {
```
GSadee:
```suggestion
            if (this.hasButtonTarget && session.isEligible('paypal')) {
```
🤔

assets/shop/controllers/PaymentPageWalletButtonController.js
```diff
@@ -20,28 +22,42 @@ export default class extends Controller {
                 scriptUrl: this.scriptUrlValue,
                 instanceConfig: this.instanceConfigValue,
                 currencyCode: this.currencyCodeValue,
+                amount: this.hasAmountValue ? this.amountValue : undefined,
```
GSadee:
```suggestion
                amount: this.amountValue,
```

GSadee (commented):
What is more:
- `payLaterEnabled` is not a part of `templates/pay_with_paypal.html.twig`, but it probably should be 🤔 
- If I'm correct, the instance of PayLater does not get the locale, but it should be

assets/shop/scripts/paypal-payment-page.js
rust-le: Fixed. Both PaymentPageCardFieldsController and PaymentPageGooglePayController now pass amount into paymentPageSession() too, so it no longer depends on which controller wins the race.

templates/shop/pay_with_paypal/content/paypal_messaging.html.twig
rust-le: Fixed, messaging now uses context.amount (payment-sourced), same as the wallet button.

templates/shop/pay_with_paypal/content/paypal_messaging.html.twig
rust-le: Applied the amount part

assets/shop/controllers/PaymentPageWalletButtonController.js
rust-le: Fixed

assets/shop/controllers/PaymentPageWalletButtonController.js
rust-le: Fixed

src/Provider/PayPalPaymentPageContextProvider.php
```diff
@@ -58,13 +60,15 @@ public function provide(PaymentInterface $payment, string $locale): array
             'errorPayPalPaymentUrl' => $this->router->generate('sylius_paypal_shop_payment_error'),
             'googlePayEnabled' => $this->fundingSourcesConfigurationProvider->isGooglePayEnabled($channel),
             'languageCode' => $this->languageCode($locale),
```
GSadee, line 62:
```suggestion
            'languageCode' => $this->languageCode($processedLocale),
```

GSadee (approved):

